Issue #34, opened 2014-11-19
SelectList won't trigger when highlighted


tomByrer:
http://jquense.github.io/react-widgets/docs/#selectlist
It is handy that you will highlight (gray background) an item when you mouse over it.  However, you can not select/click unless you click the mouse directly over the name or button/box.  The highlight in the blank area gives a false impression.
Fx35 (dev channel)

Thanks for your lib; I may use it soon!

willdady:
I agree with @tomByrer. This feels broken an unintuitive and would be better if labels were set to block by **default**. I suspect less-savvy users may be confused why nothing happens on click. I get that the fix is trivial I just think it's a more sensible default.

mattez:
I agree with @willdady.
It feels broken. Not good UX.
It should be set _display: block;_ **by default**.
Really. For sure.

PR #35, opened 2014-11-19
+ SlectList display:block trick

tomByrer:
https://github.com/jquense/react-widgets/issues/34#issuecomment-63606869


jquense:
thanks for the PR, im not gonna merge this as the docs have changeda bit since this, and I don't know that this tip is a good fit for the docs in general.

thanks for the input though!

PR #43, opened 2015-02-02
Use secure CDN urls to prevent Firefox blocking mixed active content.

ludovicofischer:
On Firefox 35, I cannot load https://jquense.github.io/react-widgets/docs/ correctly, because the mixed content protection kicks in https://developer.mozilla.org/en-US/docs/Security/MixedContent


jquense:
thanks!

PR #55, opened 2015-03-02
Highlight the current day/month/year/decade

ojame:
Our users feel a little bit lost, especially when they're flicking through a few months to have a look at dates and then come back to the current month (and I agree). Without highlighting the current day/month etc, it's very easy to lose context.

image
image
image
image

I don't really feel comfortable with the naming of `todaysDate`, but `currentDate` is already taken (though in the `state`, using the same name for a local could be confusing?).

Let me know your thoughts, thanks mate!

### DIFF
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -125,14 +125,15 @@ var Calendar = React.createClass({
     var {
         className
       , ...props } = _.omit(this.props, Object.keys(propTypes))
-      , View     = VIEW[this.state.view]
-      , unit     = this.state.view
-      
-      , disabled = this.props.disabled || this.props.readOnly
-      , date     = this.state.currentDate
-      , labelId  = this._id('_view_label')
-      , key      = this.state.view + '_' + dates[this.state.view](date)
-      , id       = this._id('_view');
+      , View       = VIEW[this.state.view]
+      , unit       = this.state.view
+
+      , disabled   = this.props.disabled || this.props.readOnly
+      , date       = this.state.currentDate
+      , todaysDate = new Date()
+      , labelId    = this._id('_view_label')
+      , key        = this.state.view + '_' + dates[this.state.view](date)
+      , id         = this._id('_view');
 
     return (
       <div {...props }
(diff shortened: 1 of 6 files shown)


jquense:
This is awesome! This has long been on the "i really need to just do that" list, so thanks. Give me a bit to test it out and we'll take it from there.

ojame:
:fist: 

Yeah, I'm unsure why the test fails, sorry mate!

This library has really come a long way, things are looking great.

jquense:
don't worry about the tests its a false positive I really need to fix :P

src/less/core.less
```diff
@@ -74,6 +74,11 @@
     width: 100%;
   }
 }
+
+.rw-now {
+  border: 1px solid #adadad;
```
jquense: I think this is the only thing i'd change, the grey border feels to much like the selected, hover, and focus styles. Not that it matters that much since its a default style. Even still i think i'd like to differentiate it a bit, heres what I thought looked ok on my end, Mostly though it makes sure not to hardcode the color so less consumers can tweak as they wish.

```less
@calendar-now-color: darken(@state-border-focus, 15%); // in bootstrap-theme.less

.rw-now {
  &,
  &.rw-state-focus {
    color: @calendar-now-color;
    font-weight: 600;
  }
}
```

No pressure to change it yourself if you don't want (let me know), but i wouldn't want to tweak it without letting you know.

src/Calendar.jsx
```diff
@@ -125,14 +125,15 @@ var Calendar = React.createClass({
...
-      , date     = this.state.currentDate
-      , labelId  = this._id('_view_label')
-      , key      = this.state.view + '_' + dates[this.state.view](date)
-      , id       = this._id('_view');
+      , View       = VIEW[this.state.view]
+      , unit       = this.state.view
+
+      , disabled   = this.props.disabled || this.props.readOnly
+      , date       = this.state.currentDate
+      , todaysDate = new Date()
```
jquense: I think `todaysDate` seems fine as a name, sorry i already took `currentDate` :P

src/less/core.less
ojame: Yep - very good suggestion, I felt the same way (but, lazy). Give me a tick to throw it in!

ojame:
@jquense I changed the `rw-now` styling, and it just bolds the string - it doesn't change the color or anything, I feel like this is good consistency especially when hover and focus states exist.

If you want to change the style of the now, that's perfectly fine :D.

Let me know if there's anything else.

jquense:
:hand: thanks!

PR #58, opened 2015-03-05
Add support for date range values to `Calendar`

dwick:
This PR adds support for date range values to the calendar widget.

Example:

```js
var start = new Date(2015, 0, 1)
   ,  end = new Date(2015, 0, 10);

<Calendar value={[start, end]} />
```

In the above example, `start` and/or `end` can be `null` as well.  If both values are `null`, then nothing is selected, and if a single value is `null` then only a single date is selected.

Let me know if you need anything else from me to get this in.

### DIFF
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -37,7 +37,10 @@ var MULTIPLIER = _.object([
 var propTypes = {
 
   onChange:      React.PropTypes.func.isRequired,
-  value:         React.PropTypes.instanceOf(Date),
+  value:         React.PropTypes.oneOfType([
+                   React.PropTypes.instanceOf(Date),
+                   React.PropTypes.array,
+                 ]),
 
   min:           React.PropTypes.instanceOf(Date),
   max:           React.PropTypes.instanceOf(Date),
@@ -77,12 +80,12 @@ var Calendar = React.createClass({
   propTypes,
 
   getInitialState: function(){
-    var value = this.inRangeValue(this.props.value);
+    var value = this.normalizeValue(this.props.value);
 
     return {
       selectedIndex: 0,
-      view:          this.props.initialView || 'month',
-      currentDate:   value ? new Date(value) : new Date()
+      view: this.props.initialView || 'month',
+      currentDateRange: value ? value : [new Date(), null]
     }
   },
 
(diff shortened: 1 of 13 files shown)


dwick:
not quite ready.  i'll reopen in a bit.

jquense:
ya can leave it open and just keep committing to the PR if that's easier for you. I'm excited to try this out, I'm not quite sure what the value is, but i'm open to finding ou :)

dwick:
@jquense I'm having a bit of trouble reasoning about the `controlledInput` mixin for this and as a result `Calendar#change` is a bit hairy. I'm also thinking drag events might be needed to do this correctly.  I've updated `example/example.jsx` with something that sort of works so you can what I'm talking about.

So far I'm pretty happy with the changes to `Month`, `Year`, `Decade`, and `Century`, but `Calendar` is pretty broken.

Would love your thoughts, not really happy with where this is headed.

dwick:
I think my use case might be too specific, closing unless you find this useful.

jquense:
hey @dwick this is cool. I think ultimately you've got a good idea here, selecting ranges of dates would be a nice feature for the Calendar. The big challenge here is getting the interaction model right. Here are some thoughts on that, if you'd like to take it on, if not we can file this away and it'll be a great starting point down the road if I get some time. Either way I appreciate you taking the time to write this up.

- the click to select model doesn't work great here (your hairy onChange). I think (like you noted) you'd need to do proper drag to select. That poses some additional issues, like needing to shift months as you dragged past the start or end of the month (think Jan 16 to Feb 5th).
- You'd also need to implement the same functionality in a keyboard only model (like ctrl + arrow key) to accommodate users that don't use a mouse.

src/Calendar.jsx
```diff
@@ -77,12 +80,12 @@ var Calendar = React.createClass({
...
   getInitialState: function(){
-    var value = this.inRangeValue(this.props.value);
+    var value = this.normalizeValue(this.props.value);
 
     return {
       selectedIndex: 0,
-      view:          this.props.initialView || 'month',
-      currentDate:   value ? new Date(value) : new Date()
+      view: this.props.initialView || 'month',
+      currentDateRange: value ? value : [new Date(), null]
```
jquense: `currentDate` determines what month/year/decade to show in any of the views, so it should always just be a single date. even if the value is a range

PR #61, opened 2015-03-11
Open multiselect popup on subsequent input focus; fixes #60

### LINKED ISSUE #60: Focus on Multiselect
multiselect

A little hard to understand with no visible clicks in the gif, but basically if you select an option via mouse click, then you have to lose focus on the input and then reclick the input to show the popup again.

I've spend a little bit looking into this to solve it for you, but the focus logic is a little hard for me to follow:

```
  _focus(focused, e){
    if (this.props.disabled === true)
      return

    this.setTimeout('focus', () => {
      if (focused)
        this.refs.input.focus()

      else
        this.refs.tagList && this.refs.tagList.clear()

      if (focused !== this.state.focused){
        focused
          ? this.open()
          : this.close();

        this.notify(focused ? 'onFocus' : 'onBlur', e)
        this.setState({ focused: focused })
      }
    })
  },
```

From what it looks like, `focused` and `this.state.focused` are both `true` when the user clicks the input the second time. I _assume_ a third var might be required?

Not sure, sorry mate!

ojame:
multifocus

As suggested, this uses the (removed) click method. Instead of click, i used the `onFocus` event, because I feel it's more warranted.

### DIFF
--- a/src/Multiselect.jsx
+++ b/src/Multiselect.jsx
@@ -180,7 +180,8 @@ var Multiselect = React.createClass({
             placeholder={this._placeholder()}
             onKeyDown={this._searchKeyDown}
             onKeyUp={this._searchgKeyUp}
-            onChange={this._typing}/>
+            onChange={this._typing}
+            onFocus={this._inputFocus}/>
         </div>
         <Popup {..._.pick(this.props, Object.keys(Popup.type.propTypes))}
           onRequestClose={this.close}>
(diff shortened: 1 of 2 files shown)


ojame:
Also - would it maybe be possible to bump the npm package after this gets merged (or soon?). :pray:

jquense:
Many Thanks! I am hoping to release tomorrow, I have a bunch of stuff floating in master I need to ensure is good, but the goal is tomorrow.

ojame:
No troubles mate, just wanted to be on the same page.

Thanks for all your hard work! :hand:

PR #62, opened 2015-03-12
Multiselect now takes a maxLength prop

ojame:
Another tiny one, heh.

Our server has a maximum length it can take for a value (for eg 255), so if tags are longer it fails. This PR allows you to pass a maxLength attribute through to cap the tag.


jquense:
+1

PR #65, opened 2015-03-16
bugfix: add missing "this" to access props

twada:
fixes "ReferenceError: props is not defined"

### DIFF
--- a/lib/DateTimePicker.js
+++ b/lib/DateTimePicker.js
@@ -317,7 +317,7 @@ var DateTimePicker = React.createClass({
 
     if (typeof format !== "function") formats.push(format);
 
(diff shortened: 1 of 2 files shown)


twada:
I've found one more undefined variable `v` here
https://github.com/jquense/react-widgets/blob/master/src/DateTimePicker.jsx#L332-L333

```javascript
    if ( this.props.parse === 'function' )
      return this.props.parse(string, v)
```

could be:

```javascript
    if ( this.props.parse === 'function' )
      return this.props.parse(string)
```

jquense:
thanks!

twada:
Thank you for merging!

PR #66, opened 2015-03-18
Fixes issue with input width in Chrome

rickharrison:
In Chrome, when you are typing in options in the multiselect, sometimes the input isn't wide enough to fit the text. By adding a slight buffer, it fixes the issue.

### DIFF
--- a/lib/MultiselectInput.js
+++ b/lib/MultiselectInput.js
@@ -23,7 +23,7 @@ module.exports = React.createClass({
   render: function () {
     var value = this.props.value,
         placeholder = this.props.placeholder,
-        size = Math.max((value || placeholder).length, 1);
+        size = Math.max((value || placeholder).length, 1) + 1;
 
     return React.createElement("input", babelHelpers._extends({}, this.props, {
       type: "text",


jquense:
Hi there,  thanks for the PR. I've never actually run into an issue where this happens could you maybe post a fiddle reproducing it? My instinct is that if that's happening then there is something weird going on with your fonts but I could be way off. 

thanks again!

rickharrison:
I can actually get this to happen on the example page: http://jquense.github.io/react-widgets/docs/#/multiselect

If I type in "Award Winning" into the first box, about halfway through the box shifts a little bit to the left and the A is cut off. In my testing, this was only happening in Chrome.

jquense:
Sorry I am not seeing it...I am running Chrome on Windows. maybe you could paset an image and provide some info about where you are running it? Maybe its an extension causing a conflict??
 
image

rickharrison:
It is not as pronounced on the docs page, but it does move while typing. It is much worse on my page, which might be a font issue.

!(http://i.imgur.com/1UsaWqz.png)

jquense:
It seems weird that Chrome would treat this differently on different platforms...The font difference makes sense to me, if you were using a wide font it might confuse Chrome. Do you think you could see if disabling extensions fixes it? 

Sorry for all the rigmarole I am always cautious about making a change to something when i don't know what the cause of the problem is. I don't want to bandaid something that needs a better solution. I do want to help ya get the problem fixed though

rickharrison:
Yea, it is unfortunately still happening with extensions disabled. In my experience, Chrome's font rendering is not the best, and custom fonts will exacerbate this issue in my case. By giving it a little extra size, it seems to alleviate causes of the problem while not causing any other.

jquense:
ok well its quite a minimal change so I'll go for it. thanks

rickharrison:
Awesome, a minor bump on npm would be sweet :)

rickharrison:
@jquense You took this change out then? Was this an accident?

jquense:
I didn't...where did it go??? O I'm sorry did you make the change directly to the `/lib` file? Those files are the compiled files, they get overridden ever release, you need to change the file in `/src`, would you mind sending another PR?

rickharrison:
Ohhhh that makes sense. I'll do another PR now.

PR #87, opened 2015-04-15
fixes #86

### LINKED ISSUE #86: ComboBox's textField prop not working with Immutable.Records
(This most likely applies to various components, not just the combobox.)

It seems the issue is caused by the [_dataText](https://github.com/jquense/react-widgets/blob/master/src/mixins/DataHelpersMixin.js#L23) function in the `DataHelpersMixin`. When checking the presence of the attribute specified by the `textField` prop, `_.has` uses `hasOwnProperty` which does not work with records:

```js
var Record = Immutable.Record({ foo: undefined }),
    r = new Record({ foo: 'bar' });

r.foo; // 'bar'
r.hasOwnProperty('foo'); // false
```

pigulla:


jquense:
Much appreciated, thanks for taking the time to do this. I think this change actually makes more sense in the accessor code ([here](https://github.com/jquense/react-widgets/blob/master/src/mixins/DataHelpersMixin.js#L13)) and not a in the `has()` helper. The reason is that other places actually depend on `has` only checking own properties, for instance I use it in `for in` loops to filter out the built in properties.

Tests can be added to the DataHelper test file as well :)

thanks!

pigulla:
Yeah, that does make sense. I'll see if I can patch something up tonight :)

PR #92, opened 2015-04-19
Unicode multiplication x

dannyrich:
Replace the "×" with "\u00d7" in the Multiselect. I was having trouble on a legacy webpage with mismatched encodings showing the wrong character (it was showing "Ã—" instead of "×").

### DIFF
--- a/lib/MultiselectTagList.js
+++ b/lib/MultiselectTagList.js
@@ -59,7 +59,7 @@ module.exports = React.createClass({
             { tabIndex: "-1", onClick: !(disabled || readonly) && _this._delete.bind(null, item),
               "aria-disabled": disabled,
               disabled: disabled },
-            "×",
+            "\u00d7",
             React.createElement(
               "span",
               { className: "rw-sr" },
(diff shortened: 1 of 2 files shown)


jquense:
Hey there @dannyrich Thanks for the PR. I'd be happy to take this but you've actually made the change in the wrong spot. the stuff in `/lib` is generated during the build so changes there will get overwritten on each new build. The spot you want is: `/src/MultiselectTagList.jsx`

dannyrich:
Ahh I see. Sorry about that. It doesn't look like a change upstream will work. One of the processors is changing the unicode string to the character, same as it did for the HTML entity. Even rewriting the return statement to not use JSX doesn't help.

I tested it on a project using [reactify](https://github.com/andreypopp/reactify) and when the JSX is parsed the unicode string is not converted to the character.

Do you know of any other processor that is being used that could be making the change? Maybe there's a flag that will prevent it from making that conversion?

jquense:
Hey there,

Babel is the cuplrit. My initial sort of checks into this have been fairly fruitless :\ Babel seems to `JSON.Stringify()` the text which is kill the unicode string. I am not sure how best to articulate the problem with this, so you may want to submit an issue here, they are usually really responsive. https://github.com/babel/babel/issues

PR #114, opened 2015-05-06
Add bower.json to .npmignore

Morhaus:
See https://github.com/webpack/webpack/issues/1042

This ensures webpack will not use bower.json when react-widgets is installed from npm.


jquense:
thanks!

PR #142, opened 2015-06-17
Changed font-related LESS variables prefix from "fa" to "rw" to avoid clash with Font-Awesome

mathieumg:
If a project imports LESS styles from Font-Awesome and then also imports LESS styles from React-Widgets, there is a clash even though the font files are different. I imagine you started from a piece of code you borrowed from Font-Awesome but never bothered to change the name of the variables, you have a good reason now! :grin:

### DIFF
--- a/lib/less/icons.less
+++ b/lib/less/icons.less
@@ -1,17 +1,17 @@
 
 @font-face {
   font-family: 'RwWidgets';
-  src: ~"url('@{fa-font-path}/rw-widgets.eot?v=@{fa-version}')";
-  src: ~"url('@{fa-font-path}/rw-widgets.eot?#iefix&v=@{fa-version}') format('embedded-opentype')",
-    ~"url('@{fa-font-path}/rw-widgets.woff?v=@{fa-version}') format('woff')",
-    ~"url('@{fa-font-path}/rw-widgets.ttf?v=@{fa-version}') format('truetype')",
-    ~"url('@{fa-font-path}/rw-widgets.svg?v=@{fa-version}#fontawesomeregular') format('svg')";
+  src: ~"url('@{rw-font-path}/rw-widgets.eot?v=@{rw-version}')";
+  src: ~"url('@{rw-font-path}/rw-widgets.eot?#iefix&v=@{rw-version}') format('embedded-opentype')",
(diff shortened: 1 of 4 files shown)


dozoisch:
:+1:

jquense:
hey! thanks for the PR :) much appreciated, I am still trying to decide if this constitutes a breaking change or not. I do have a question though, I have kept the variable name the same for now b/c it is actually jsut a subset of FA, and its easier to update if the variables stay the same. I have no problem changing it tho if its conflicting with other usage of FA. so my question: when does this conflict, or rather how are you using the files that they conflict? Just interested for my own sake :)

thanks again!

mathieumg:
If it's a subset, the same codes are used for the same icons and it will always stay that way, perhaps the  `font-family` and the font file name should be made less variables too? That way, it could be set so that React-Widgets leverages the same font file as the current Font-Awesome installation. That is not absolutely necessary however.

I would imagine this constitutes a breaking change for anyone using the LESS files instead of the built CSS.

This approach conflicts when doing the following, for example:

```less
@import        "~bootstrap/less/bootstrap"; // Imports the LESS from Bootstrap.
@import        "bootstrap-theme"; // Overrides some rules from stock Bootstrap to define a theme.

@import        "~font-awesome/less/font-awesome"; // Imports the LESS from Font-Awesome.
@fa-font-path: "~font-awesome/fonts"; // Overrides where Font-Awesome fonts are located.

@import       "~react-widgets/lib/less/react-widgets"; // This overrides fa-font-path and causes issues with Font-Awesome because LESS variables propagate "upwards".

@import       "mystyles"; // In-house styling rules.
```

The use-case is: leverage our Bootstrap theme for the React-Widgets components as well.

I denoted another issue with regards to the styling of React-Widgets components when using that approach, but let's kill one pigeon at a time.

mathieumg:
@jquense Were you waiting on my for something, or is there anything else I can do to help this issue progress? Thanks!

jquense:
Hey sorry, the PR looks great, I just need to figure out if I am considering it a breaking change, and just haven't had a moment to think it through yet :)

mathieumg:
No worries, thanks!

PR #149, opened 2015-07-01
add editorconfig file

AlexKVal:


jquense:
I am going to be honest here. I have no idea what an editor config is...

mathieumg:
> I am going to be honest here. I have no idea what an editor config is...

I don't personally use it (lack of time to set it up properly, I imagine), but it's a config file that describes what "file style" (tabbings, carriage returns, etc.) a project uses, so contributors with a compatible editor can edit any source file from the project without messing up the current file style. See: http://editorconfig.org/

AlexKVal:
It is a convenience for those developers, who maintains several projects with different formatting preferences.
Mainly 'tab' vs 'spaces'.
Editor (with `editorconfig` plugin) helps to deal with it automatically.
It is very convinient :+1: :wink:

PR #163, opened 2015-07-10
Leverage parser for passing strings as date value

aleemb:
This makes it easier to pass in strings as the date `value` and `defaultValue` instead of limiting these to a date object. This is useful since passing in a string from the model is likely the most common use-case.

A new prop called `valueFormat` is also exposed to specify the format for the `value` and `defaultValue`. Although the `editFormat` and `format` props are also tried for parsing the date string making `defaultValue` optional.

### DIFF
--- a/lib/DateTimePicker.js
+++ b/lib/DateTimePicker.js
@@ -123,7 +123,8 @@ var DateTimePicker = React.createClass({
     var dateListID = this._id('_cal');
     var dropUp = this.props.dropUp;
     var renderPopup = _.isFirstFocusedRender(this) || this.props.open;
-    var value = dateOrNull(this.props.value, [this.props.valueFormat, this.props.editFormat, this.props.format], this.props.culture);
+    var dt = this.props.value;
+    var value = dateOrNull((typeof dt === 'string') ? this._parse(dt) : dt);
     var owns;
 
     if (dateListID && this.props.calendar) owns = dateListID;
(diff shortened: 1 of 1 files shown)


jquense:
Hey thanks for this. I am still a bit unsure if this is a good idea to put in the library but since we already parse strings to dates it might be worth it.

A few things before I can merge this:
- rather then add a new prop, lets just use the existing ones and the `_parse()` method that is already there, that will keep a nice symmetry in the string parsing logic.
- could you please add a test or two around this feature
- update the docs

aleemb:
> since we already parse strings to dates it might be worth it

That was my thinking too (plus the convenience). However, I have been playing with the Date API and discovered the `Date.parse` (IE10+) which makes this fairly trivial:

    value={ new Date( Date.parse(this.state.somedate) ) }

Maybe the same logic can go in the `dateOrNull` function:

    if (typeof dt === 'string' && Date.parse) dt = Date.parse(dt);

This makes it simpler and relies on the browser implementation. Thoughts?

jquense:
Since we support ie8+ that probably isn't going to work. it also also fairly prone to failure when dealing with real servers, like Microsoft's inane date format: `"\/Date(1320825600000-0800)\/"`. I definitely see the value in making the input more "native" but in this case its probably better to just keep value input and typing input consistent with something like,: `dateOrNull(typeof dt === 'string' ? this._parse(dt) : dt)`

aleemb:
@jquense without the dateOrNull calls in `_selectDate` I see the following error when selecting a date from the dropdown:

    Uncaught TypeError: date[("get" + method)] is not a function

The stack trace points to the following lines:

```
_selectDate:
dateTime = dates.merge(date, dt),

_notify:
this.props[handler] && this.props[handler].apply(null, [].concat(args));    

_change:
if (this.props.onChange && this.state.view === this.props.initialView) return this.notify('onChange', date);
```

aleemb:
Please feel free to disregard this PR if it looks off. Not sure if the code is rightly in order.

jole78:
why was this closed and is there a way to get this working??
I have some dates coming from the server as json "2016-11-04T11:33:20.453670Z" and now I can't get that to work with the DateTime widget.

Issue #164, opened 2015-07-16
Integration with RequireJS


cosminnicula:
Although I've configured the shims so that `ReactWidgets` is loaded after `React`, I do get an error `Uncaught TypeError: Cannot read property 'version' of undefined` as if React is undefined.

Any hint on this?

**index.html**
```
<!DOCTYPE html>
<html>
	<head>
		<title>ReactWidgets with RequireJS</title>
	</head>
	<body>
		<div id="dropdown-list-placeholder"></div>
		<script data-main="main" src="../../node_modules/requirejs/require.js"></script>
	</body>
</html>
```

**main.js**
```
require.config({
	paths: {
		'jQuery': '../../node_modules/jquery/dist/jquery.min',
		'underscore': '../../node_modules/underscore/underscore-min',
		'backbone': '../../node_modules/backbone/backbone-min',
		'consolePolyfill': '../../node_modules/console-polyfill/index',
		'react': '../../node_modules/react/dist/react.min',
		'reactWidgets': '../../node_modules/react-widgets/dist/react-widgets'
	},
	shim: {
		'jQuery': {
			exports: '$'
		},
		'underscore': {
			exports: '_'
		},
		'backbone': {
			deps: ['underscore', 'jQuery', 'consolePolyfill'],
			exports: 'Backbone'
		},
		'reactWidgets': {
			deps: ['react']
		}
	}
});


require(['react', 'reactWidgets', 'backbone'], function() {

	var MainSectionView = Backbone.View.extend({

		render: function() {

			React.render(new ReactWidgets.DropdownList({data: ['orange', 'red', 'blue', 'purple']}), this.$('#dropdown-list-placeholder').get(0));

			return this;
		}
	});

	new MainSectionView().render();
});
```

jquense:
Unfortunately I don't know anything about requireJS :/ so its really hard for me to help with this, but it does look like maybe you forgot the Globalize.js dependency

cosminnicula:
Ok, then let's forget about RequireJS.

Consider a simple html page that needs to render a DropdownList. This still gives a nasty error: `Uncaught Error: Invariant Violation: React.render(): Invalid component element. This may be caused by unintentionally loading two independent copies of React.`

What I am missing here?

**index.html**
```
<html>
	<head>
		<title>ReactWidgets</title>
	</head>
	<body>
		<div id="dropdown-list-placeholder"></div>

		<script src="../../node_modules/jquery/dist/jquery.min.js"></script>
[...]

PR #165, opened 2015-07-17
Export the library as UMD format

cosminnicula:
Support for AMD, CommonJS2 and global variable. Related with #164

### DIFF
--- a/webpack.configs.js
+++ b/webpack.configs.js
@@ -123,7 +123,12 @@ module.exports = {
 
     externals: {
       globalize: 'Globalize',
-      react:  'React'
+      react: {
+	    root: 'React',
+	    commonjs: 'react',
+	    commonjs2: 'react',
+	    amd: 'react'
+	  }
     }
   }),
 
--- a/webpack.configs.js
+++ b/webpack.configs.js
@@ -118,7 +118,7 @@ module.exports = {
       path: path.join(__dirname, './dist'),
       filename: 'react-widgets.js',
       library:  'ReactWidgets',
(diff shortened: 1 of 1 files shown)


cosminnicula:
This should also be published on NPM.

jquense:
Can you walk me through why you need this? The npm package is already in CommonJS format, why not just consume the modules instead of the dist build?

cosminnicula:
The use case is documented in #164. It's mostly because other projects / applications might already consume AMD-format packages or global variables. By limiting the package to be consumed only as CommonJS, you exclude those type of applications.

Other libraries already have umd as a default output:
- Facebook's FixedDataTable https://github.com/facebook/fixed-data-table/blob/master/webpack.config.js
- React Router https://github.com/rackt/react-router/blob/master/webpack.config.js
- React Bootstrap https://github.com/react-bootstrap/react-bootstrap/blob/master/webpack/webpack.config.js

Please merge both commits in this PR before publishing to NPM.

jquense:
merging since its easy enough for me to provide this, but do want to say that, using the dist/ build is generally not recommended for a variety of reasons, that make it less useful then using something like browserify or webpack, but you are certainly welcome to it if thats what you need :)

cosminnicula:
Yes, I know and you're right.
But there are projects out there that are still relying on AMD format or global variables and do their build with Maven. You cannot force them to consume a library as CommonJS or use browserify or webpack. We should care about the poor :)

cosminnicula:
Can you publish this on NPM? Thanks!

PR #192, opened 2015-09-07
onChange source argument to distinguish between date-change and time-…

pvsrivathsa:
…change


pvsrivathsa:
None of the reasons for failure is related to the changes that I've done

jquense:
Hey there,

thanks for the PR, sorry its taken so long to get back to you. The one problem with this approach is that it doesn't cover the case of typing a date into the input. I would think that this argument probably belongs more on the `onSelect` event than the `onChange` one

PR #217, opened 2015-10-01
fix mobile device that use touch events

danfernand:
Currently broken in mobile devices (broswers) because the onClick event
in SelectInput is not firing. A current workaround would be to loose
focus from the component and then re-focus to gain the popup list.

Tested and broken in
----------------------------
ipad air, ipad 2

I'm sure other mobile devices that have touch events enabled are broken
as well


jquense:
thanks!

PR #220, opened 2015-10-03
Fix linting errors

conorhastings:
@jquense

### DIFF
--- a/src/Multiselect.jsx
+++ b/src/Multiselect.jsx
@@ -120,12 +120,12 @@ var Multiselect = React.createClass({
   getInitialState(){
     var { data, value, valueField, searchTerm } = this.props
       , dataItems = splat(value).map( item => dataItem(data, item, valueField))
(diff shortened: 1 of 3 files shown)


src/Multiselect.jsx
```diff
@@ -118,9 +118,9 @@ var Multiselect = React.createClass({
   },
 
   getInitialState(){
-    var { data, value, valueField, searchTerm } = this.props
+    var { value, valueField, searchTerm } = this.props
       , dataItems = splat(value).map( item => dataItem(data, item, valueField))
-      , data = this.process(data, dataItems, searchTerm)
+      , data = this.process(this.props.data, dataItems, searchTerm)
```
jquense: this is gonna break here, `data` is used on the line above, i'd keep it and use processedData as the new var

conorhastings:
@jquense cool, good catch, updated.

jquense:
thanks

PR #221, opened 2015-10-03
cancel mouseDown of number picker onMouseOut

### LINKED ISSUE #213: bug when mouse is pressed, but leaves the element
Using Chrome i could reproduce a strange behavior in the demo numberpicker as well.
When you press the "+" or "-" Buttons with your mouse and keep the mouse pressed, but leave the element (so it's not hovered anymore), the numberpicker won't stop to increase/decrease.
I guess it's a problem with the mouseover cancel event.

conorhastings:
closes #213 @jquense

### DIFF
--- a/src/NumberPicker.jsx
+++ b/src/NumberPicker.jsx
@@ -114,6 +114,7 @@ var NumberPicker = React.createClass({
             className={cx({ 'rw-state-active': this.state.active === directions.UP})}
             onMouseDown={this._mouseDown.bind(null, directions.UP)}
             onMouseUp={this._mouseUp.bind(null, directions.UP)}
+            onMouseOut={this._mouseUp.bind(null, directions.UP)}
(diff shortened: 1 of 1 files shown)


jquense:
:+1: many thanks

conorhastings:
happy to contribute :grin: just saw you're in washington dc, you should check out the react meetup a few people from my company run if you don't attend already , http://www.meetup.com/React-DC/events/225703899/

jquense:
I actually just moved up from DC to NJ (need to update my info). I've been to a few of the react DC meet ups tho, ya'll do a good job!

conorhastings:
ahhh too bad, that's cool though. Would you mind publishing these recent changes that have gotten in? we're using this heavily in production and would love some of these fixes.

jquense:
done! 2.8.2 is out

PR #245, opened 2015-11-06
Show previous days in calendar when min limit is set

tomaszkepa:
In my opinion looks better, no empty space in calendar :)


jquense:
Thanks for the PR, sorry it's taken so long to get back to you here. I think I'd like to solve this in a way that allows for both approaches, and doesn't change the default.

I am going to close this for now but if you'd like to submit a PR that adds something like an `emptyDayComponent` prop that by default renders nothing (but passes in the day so you can customize it yourself) I'd be happy to take it/

PR #258, opened 2015-11-26
NumberPicker updated to respond to onMouseLeave rather than onMouseOut.

benwiley4000:
Very simple change/fix. Basically, the NumberPicker's up/down caret buttons run `this._mouseUp` in response to React's `onMouseOut`, but we should be using `onMouseLeave` instead. `onMouseOut` is triggered whenever the mouse exits the bounds of any child element inside the targeted element, so when the mouse goes from the up or down caret to the button background, numeric scrolling stops. The only change I made to the source files was replacing the two instances of `onMouseOut` with `onMouseLeave`. I built and tested; with this change, numeric scrolling only stops once the button is exited, as desired.

Note: All I did was make the two minor changes, then `npm install`, and `npm run build`. I'm not sure why this effected changes in so many files but I'm hoping that's what was supposed to happen.

### DIFF
--- a/src/NumberPicker.jsx
+++ b/src/NumberPicker.jsx
@@ -115,7 +115,7 @@ let NumberPicker = React.createClass({
             className={cx({ 'rw-state-active': this.state.active === directions.UP})}
             onMouseDown={this._mouseDown.bind(null, directions.UP)}
             onMouseUp={this._mouseUp.bind(null, directions.UP)}
-            onMouseOut={this._mouseUp.bind(null, directions.UP)}
+            onMouseLeave={this._mouseUp.bind(null, directions.UP)}
             onClick={this._focus.bind(null, true)}
             disabled={val === this.props.max || this.props.disabled}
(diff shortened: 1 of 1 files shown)


benwiley4000:
p.s. Not sure what's up with this failed test, but the module definitely worked for me after I made the changes.

jquense:
the CI is always failing nowadays don't worry too much about it. would you mind cleaning up the PR tho to just include the src changes? 

looks good otherwise I think! thanks for the PR

benwiley4000:
Oh yeah sure. I'll just make a new request, and no problem! :)

benwiley4000:
Just kidding about making a new request. This one has been updated! Should be ready to go.

benwiley4000:
Bumping - just wanted to make sure you saw I had fixed the request.

jquense:
thanks!

benwiley4000:
Thank you! Do you think you might have a new release pushed to npm soon?

PR #271, opened 2015-12-17
Add default browserify transform to package.json

GeyseR:
This will help preventing additional transformation over resulting files, produced by browserify.
Reactjs works in same way, see https://github.com/facebook/react/blob/master/packages/react/package.json#L29
One difference, that i added envify to devDependencies section

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "chance": "^0.7.6",
     "cpy": "^3.4.0",
     "css-loader": "^0.15.6",
+    "envify": "^3.0.0",
     "eslint": "^0.24.1",
     "eslint-plugin-mocha": "^0.5.1",
     "eslint-plugin-react": "^2.7.1",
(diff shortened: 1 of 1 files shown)


jquense:
hey , thanks for the PR, sorry I haven't used browserify in a while so this is a bit opaque to me. Does this work recursively? Meaning, if someone requires this from npm and then uses browserify for itself it will check the local package.json for module specific transforms?

My assumption was this is only helpful when bundling just this module with browserify, which is already handled by the "dist" bundle (using webpack)

GeyseR:
Hi, @jquense!
>Does this work recursively?
Yes. Moreover, if local js modules import react-widgets from node_modules, then regular transformations will be applied only for local js code. So, developer have to add additional step to his build process with global transformation to post-process whole result file. 
You can read detail on https://github.com/substack/module-deps#transforms docs page

PR #288, opened 2016-01-13
Add missing packages that gulp requires

lemiii:
When running gulp, it has errors stating that packages are missing. package.json has been updated to allow NPM to install them.


jquense:
gulp isn't actually used anymore for doing anything, what are you trying to do?

lemiii:
I'm trying to change the rw-i-* icons to use fa-*

jquense:
look here: https://github.com/jquense/react-widgets/blob/master/package.json#L29-L52

Those are all the commands used to build everything. to build everything just run: `npm run build`

and thanks for reminding me I need to delete the gulpfile

lemiii:
Thanks. I'm so used to building with gulp. I hadn't realized NPM could run builds as well.

PR #291, opened 2016-01-14
add instruction to run test for development

goldensunliu:


jquense:
hi there thanks for the PR, You might want to also add a note about `npm run tdd` which will put it in watch mode

PR #297, opened 2016-01-19
remove isRequired from NumberInput parse prop to prevent NumberPicker propType warning

### LINKED ISSUE #293: Required prop `parse` was not specified in `NumberPickerInput`
Not sure why I'm getting this error while attempting to use NumberPicker. Not sure if this is related to globalize (as it's been the cause of problems for me lately), localization is setup as follows.

```
var Globalize = require('globalize');
var globalizeLocalizer = require('react-widgets/lib/localizers/globalize.js');
Globalize.locale('en');
globalizeLocalizer(Globalize);
```

Skimming over the source code for NumberPicker it looks like it does in-fact pass `this.props.parse` to NumberInput which would explain the error. Does that mean providing a **parse** prop to NumberPicker is required? The documentation doesn't say it's a required field. I'm a bit stuck here.

conorhastings:
NumberPicker takes an optional parse function but since it is required in NumberInput, you receieve a warning if you do not provide a prop. parse was removed from getDefaultProps in 3.1.5 , this removes the isRequired for parse from the PropTypes declaration in NumberInput to avoid this warning

closes #293

### DIFF
--- a/src/NumberInput.jsx
+++ b/src/NumberInput.jsx
@@ -14,7 +14,7 @@ export default React.createClass({
 
     format:       CustomPropTypes.numberFormat,
 
-    parse:        React.PropTypes.func.isRequired,
+    parse:        React.PropTypes.func,
     culture:      React.PropTypes.string,
 
     min:          React.PropTypes.number,


conorhastings:
@jquense thanks! is there any way you could release this in a patch version?

jquense:
released!

conorhastings:
awesome, thank you!

Issue #294, opened 2016-01-18
Open the calendar to a different date


yleclanche:
Hello, 

I don't find the appropriate props in the documentation, so I'll ask here : 
When you open the calendar, it shows the current month. What if I want to show a different month by default ? Is it possible to open the calendar on another month ?

Thanks.

jquense:
Hi there,

The Calendar will open to whatever the `value` of the Calendar/Datepicker is, when no value it defaults to the current month. Currently though there is no way of adjusting that default.

yleclanche:
Thanks for your answer.

slybridges:
Hi @jquense,

Would you be willing to reopen this issue if I would submit a PR that would allow to open the calendar at a specific date ?

I'm thinking for example of adding a props `defaultDate` to `Calendar` and `DateTimePicker` components that could optionally take a `Date object`. If none are provided, it would revert to the current behaviour using today's date.

Let me know your thoughts, i really would need to have that feature on my project :-)

jquense:
I might be open to it but can you talk to me about what the use case is?

slybridges:
I have a travel form. A travel has many stops, each stop has a departure and arrival date). 
The dates are optional but they must be in a chronological order. 

When the user selects a date, I want the DatePicker to open with the view of the previously selected date in the form (i.e. if i'm editing the arrival of the 3rd stop in the form and no value has been selected yet, I want the default view to show the departure date of the 2nd stop by default).

In my case the dates are optional so I don't want to use `defaultValue` or pre-fill input box, I just want to have the DatePicker open on a specific date so that the user doesn't waste clicks in getting back close to the date he just entered.

jquense:
that seems fair, though I think the correct way to do this is probably: #261 which means pushing the calendar state "currentDate" out to props, then it can take advantage of the uncontrolled bahvior, and have a defaultCurrentDate

jquense:
you can see I've already started to mock out the API here: https://github.com/jquense/react-widgets/blob/master/src/Calendar.jsx#L434 if you want to pick it up and finish it i'd happily merge it

PR #300, opened 2016-01-20
Expose currentDate to props. Fixes #261 and #294.

### LINKED ISSUE #261: [Feature request] Calendar: Display a date without setting value
To display a specific date in the calendar widget I have to select it (by setting the "value" prop). I want to set & update the displayed dates without having to select a date. Ideally, I would like to be able to set a value, but display another.

My current use case:

I have two linked calendars that should display two consecutive months. Navigating in one updates the other, using onNavigate to update the values. I would like to do this without having to actually select a date in each calendar.

image

slybridges:
Hi @jquense - let me know any comment.

I kept `currentDate` in the state to minimize the changes and because I wasn't sure on how to make it work with `uncontrollable` and how to update all the tests (the few tries I made weren't conclusive)

### DIFF
--- a/docs/components/pages/Calendar.api.md
+++ b/docs/components/pages/Calendar.api.md
@@ -18,10 +18,17 @@ Callback fired when the Calendar navigates between views, or forward and backwar
 
 The minimum date that the Calendar can navigate from.
 
-
 ### max?{ type: 'Date' }
 
 The maximum date that the Calendar can navigate to.
+
+### currentDate?{ type: 'Date', default: 'Date()', handler: 'onCurrentDateChange', controllable: true }
+
+Default current date at which the calendar opens. If none is provided, opens at today's date or the `value` date (if any).
+
+### onCurrentDateChange?{ type: 'Function( Date? date )' }
+
+Change event Handler that is called when the currentDate is changed. The handler is called with the currentDate object
 
 ### footer?{ type: 'Boolean', default: 'false' }
 
@@ -124,6 +131,7 @@ title and screen reader text for the left arrow button
 ### messages.moveForward?{ type: 'String', default: '"navigate forward"' }
 
 title and screen reader text for the right arrow button
+
 
 ## Keyboard Navigation
 
--- a/src/Calendar.jsx
+++ b/src/Calendar.jsx
@@ -65,6 +65,9 @@ let propTypes = {
   min:           React.PropTypes.instanceOf(Date),
   max:           React.PropTypes.instanceOf(Date),
 
+  currentDate:         React.PropTypes.instanceOf(Date),
+  onCurrentDateChange: React.PropTypes.func,
+
   initialView:   React.PropTypes.oneOf(VIEW_OPTIONS),
 
   finalView(props, propname, componentName){
@@ -117,12 +120,9 @@ let Calendar = React.createClass({
   propTypes,
 
   getInitialState(){
-    var value = this.inRangeValue(this.props.value);
-
     return {
       selectedIndex: 0,
-      view:          this.props.initialView || 'month',
-      currentDate:   value ? new Date(value) : this.inRangeValue(new Date())
+      view:          this.props.initialView || 'month'
     }
   },
 
@@ -132,6 +132,7 @@ let Calendar = React.createClass({
       value:        null,
       min:          new Date(1900, 0, 1),
       max:          new Date(2099, 11, 31),
+      currentDate:  new Date(),
 
       initialView:  'month',
       finalView:    'century',
@@ -142,6 +143,10 @@ let Calendar = React.createClass({
       ariaActiveDescendantKey: 'calendar',
       messages: msgs({})
     }
+  },
+
+  componentWillMount() {
+    this.changeCurrentDate(this.props.value)
   },
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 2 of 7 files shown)


src/Calendar.jsx
```diff
@@ -153,7 +157,7 @@ let Calendar = React.createClass({
     //if the value changes reset views to the new one
     if ( !dates.eq(val, dateOrNull(this.props.value), VIEW_UNIT[view]))
       this.setState({
-        currentDate: val ? new Date(val) : new Date()
+        currentDate: this.getDateOrCurrent(val)
```
jquense: So the way uncontrollable works is that the handler/prop pair are managed for you like they might be if you `setState()`

In this case you have `onCurrentDateChange` and `currentDate` when you want to "set" current date, instead of setting the state, pass that value to `this.props.onCurrentDateChange`. You can look at how `value` and `onChange` work here.

src/Calendar.jsx
```diff
@@ -168,7 +172,7 @@ let Calendar = React.createClass({
 
     var View = VIEW[view]
       , unit = VIEW_UNIT[view]
-      , todaysDate = new Date()
+      , todaysDate = this.getDateOrCurrent(null)
```
jquense: this actually should be today's date, always.

src/Calendar.jsx
slybridges: That's what started to do.
But then many tests using `BaseCalendar` were failing because they couldn't find `this.props.onCurrentDateChange`. 
How to access the wrapped container state and methods if we replace `BaseCalendar` by `Calendar` ?

src/Calendar.jsx
slybridges: I was thinking that when we set a currentDate through props, we want this to be the default value to pick. Hence we would want to replace today's date as well. 

At least, that's what worked the best from a UX perspective in my use case :-)

But happy replace it back to today's date if you think that's the way to go.

src/Calendar.jsx
jquense: the Wrapped Component will inject onCurrentDateChange, but the Base one won't. Since the DatePicker uses the Plain ol Calendar the easiest way to handle that is to include currentDate/onCurrentDateChange here as well: https://github.com/jquense/react-widgets/blob/master/src/DateTimePicker.jsx#L464

slybridges:
Hi @jquense - Updated after your comments, thanks !

For 2 of the tests I had to make some changes that maybe could be improved:
- min/max constraint test: I accessed the wrapped component props through `picker.refs.inner.props`
- culture test: `findTag` returned an error as there was 2 `theads` in the rendered component (not sure why... it displays properly with only one `thead` if I render that component in a page) so I used `scryRenderedDOMComponentsWithTag`instead.

src/Calendar.jsx
```diff
@@ -152,9 +157,7 @@ let Calendar = React.createClass({
 
     //if the value changes reset views to the new one
     if ( !dates.eq(val, dateOrNull(this.props.value), VIEW_UNIT[view]))
-      this.setState({
-        currentDate: val ? new Date(val) : new Date()
-      })
+      this.changeCurrentDate(val)
```
jquense: I am a bit concerned about this causing update thrash. Since the common case is that the Calendar will

- trigger a `value` change
- receive new `value`
- trigger a `currentDate` change
- receive new `currentDate`

It might make more sense to trigger a currentValue change at the same time as the value change [here](https://github.com/jquense/react-widgets/blob/master/src/Calendar.jsx#L296). The problem though is you still need some version of the above check in case a new value is set from outside. 

This might just be the option for now...I wonder if it makes sense to remove this behavior entirely though

src/Calendar.jsx
```diff
@@ -304,6 +308,12 @@ let Calendar = React.createClass({
   },
 
   @widgetEditable
+  changeCurrentDate(date){
+    var inRangeDate = this.inRangeValue(date ? new Date(date) : this.props.currentDate)
+    notify(this.props.onCurrentDateChange, inRangeDate)
```
jquense: I would add an equality check here, so you don't trigger an update when the value hasn't changed in a relevant way, probably: `dates.eq(inRangeDate, dateOrNull(this.props.currentDate), VIEW_UNIT[view])`

slybridges:
@jquense - Updated after your comments.

slybridges:
Hi @jquense - Anything you need from me to get this merged ?

jquense:
I don't think so, it looks good I but I need a moment to try it all out first :)

slybridges:
Hi @jquense - I realized that `currentDate` didn't apply to default time selection (that would still use today's current time).
I just fixed it (along with some linting errors that were preventing running the tests).

docs/components/pages/Calendar.api.md
```diff
@@ -18,11 +18,18 @@ Callback fired when the Calendar navigates between views, or forward and backwar
 
 The minimum date that the Calendar can navigate from.
 
-
 ### max?{ type: 'Date' }
 
 The maximum date that the Calendar can navigate to.
 
+### currentDate?{ type: 'Date', default: 'Date()' }
```
jquense: this should be: 

`currentDate?{ type: 'Date', default: 'Date()', handler: 'onCurrentDateChange', controllable: true }`

jquense:
Sorry for the delay on this! can you make that one change, and clean up and squash the commits?

thanks!

slybridges:
Hi @jquense, Doc updated and commits squashed :-)

jquense:
Thanks a lot, sorry for the drag on it much appreciated :)

PR #305, opened 2016-01-27
Fix autoFocus

jamesknelson:
As `autoFocus` is listed in `propTypes`, the `_.omit` call on line 95 was removing it from `this.props` and thus causing the variable to always be `null`.

I've tested this by hand in a project I'm using NumberPicker in. I've spent some time trying to write a test, but given that React implements `autoFocus` itself instead of passing it to the underlying HTML element, I couldn't figure out how to do this.


jquense:
thanks!

PR #317, opened 2016-02-09
Make DropdownList button consistent with other widgets

smashercosmo:


jquense:
hey there, thanks for the PR

The reason its a span in this case is for a11y, the entire wrapper of the dropdown is the "button" and the icon, is just a label really. Unlike the other pickers where the select functions as a button.

smashercosmo:
Got it) Thx for explanation.

smashercosmo:
May be worth providing a comment in a code to avoid confusion in future?

PR #318, opened 2016-02-09
disableKeyboard feature added

cyberman72:
Hello, I like your components.
I found that when the calendar is opened is not possible to change the text with keyboard because arrows are used to move the current day.
I think that will be useful to permit the developer to choose if, for the user, the keyboard is used to move betweeb days or to edit date.
Then I add a line of code to return if the property is specified.
Thank you if you implement it.

### DIFF
--- a/src/DateTimePicker.jsx
+++ b/src/DateTimePicker.jsx
@@ -318,11 +318,14 @@ var DateTimePicker = React.createClass({
 
   @widgetEditable
   _keyDown(e){
-    let { open, calendar, time } = this.props;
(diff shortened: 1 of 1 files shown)


jquense:
thanks for the PR. You can actually already accomplish this if you want. notice this line: https://github.com/jquense/react-widgets/pull/318/files#diff-9fb5fb34bac0143f00d57b9b184a8bddL325.

if the event is prevented then none of the keyboard logic is run so you can just add an `onKeyDown` handler and call `e.preventDefault()` on the event.

PR #323, opened 2016-02-15
Handle invalid dates in the moment localizer

evax:
This fixes an issue where an invalid date would trigger an invariant violation.


jquense:
thanks!

Issue #330, opened 2016-02-25
[DropdownList] missing bottom border


nmaves:
Strange but it seems to only happen on the first filtered dropdown on the page. I can reproduce this with my own local application as well.  Not sure if this is a browser issue but I get the same results in both Chrome and Safari on a Mac.

Even on your demo site you can see this 
![react_widgets](https://cloud.githubusercontent.com/assets/367750/13325854/d3b4014a-dba0-11e5-8df4-6cb0dc92da22.png)

But the second one on the page has the bottom border.

![react_widgets](https://cloud.githubusercontent.com/assets/367750/13325879/f65028a0-dba0-11e5-948a-5a980f82aaf7.png)

jquense:
I think its probably do to pixel rounding/snapping for `em` sizes :/ Feel free to PR something if you want :)

nmaves:
Any ideas on where to start?

jquense:
well that input has a borderless input inside of a div styled to look like an input (to add the icon) It seems at some line-height/font-sizes the height of the  inner input is calculating larger then its wrapper. I'd check on the height settings and em calculations there I think

nmaves:
Any thoughts on my quick fix?

MakarovAV:
Still broken for me atm (version 3.4.6).

PR #331, opened 2016-02-25
Added a set height to the .rw-filter-input

nmaves:
#330 Possible fix


jquense:
hey there! really sorry about the delayed response. I think this is fine, but I'd really want to express the heights in terms of each other, so that there isn't any special logic when you want to adjust the `@input-height` variable...I am now actually wondering how other folks handle these cases...

PR #332, opened 2016-03-01
Add support for array data items in Dropdown and ComboBox

ratheous:
I wanted to use your DropdownList and ComboBox components, but I didn't want to change my internal options structure, which was an array of arrays, where the first element of each item is the value and the second is the label. 

For example:
```javascript
var data = [
      [ 1, 'jimmy' ],
      [ 2, 'sally' ],
      [ 3, 'pat' ]
    ];

var instance = render(
          <Dropdown defaultValue={1} data={data} valueField={0} textField={1} />);
```

Your code _almost_ supported this already, if you make the values of valueField and textField strings, but not quite. This PR includes support for the above, including tests (as far as I got on them).

As it turns out, I ended up removing the components because my designer rebelled at your use of inline styles, but I had already made these changes and I thought others might find them useful. Feel free to merge if you like.

Thanks!

### DIFF
--- a/src/util/dataHelpers.js
+++ b/src/util/dataHelpers.js
@@ -7,14 +7,14 @@ function accessor(data, field){
     value = field(data)
   else if (data == null)
     value = data
-  else if (typeof field === 'string' && typeof data === 'object' && field in data)
+  else if ((typeof field === 'string' || typeof field === 'number') && typeof data === 'object' && field in data)
     value = data[field]
 
   return value
 }
 
 export function dataValue(item, valueField){
-  return valueField && item && has(item, valueField)
+  return valueField !== undefined && valueField !== null && item && has(item, valueField)
     ? item[valueField]
     : item
 }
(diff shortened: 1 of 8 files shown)


src/util/dataHelpers.js
```diff
@@ -50,10 +50,10 @@ export function dataItem(data, item, valueField){
   // make an attempt to see if we were passed in dataItem vs just a valueField value
   // either an object with the right prop, or a primitive
   // { valueField: 5 } || "hello" [ "hello" ]
-  if (has(item, valueField) || typeof first === typeof val)
```
ratheous: Just realized this slipped in to the commit. A variable named 'val' does not seem to exist, so I took a guess that it was supposed to be 'item'. I couldn't tell that it made a difference either way.

jquense:
inline styles? I wondering what you mean by that, we are like the only react library that doesn't use inline styles :P

thanks, I'm not usre this a great fit, but I'll think a bit about it. Mostly tho I think it would have worked with arrays if you just used string numerical keys as the valueField and textField (`valueField='0'`)

ratheous:
Well, at a glance I see:
```html
<div style="display: none; overflow: hidden; height: 70px;" class="rw-popup-container rw-popup-animating" data-reactid="foo">
<div style="transform: translateY(-100%); position: absolute;" class="rw-popup rw-widget" data-reactid="bar">
```

Perhaps that's created by a JS animation library, but I didn't make the decision in any case. My designer didn't want to try to customize it in the time we had available, so there you go.

As I said about the feature, it almost worked with numerical keys as text, but not. You can see for yourself if you test it. Regardless, feel free to merge or not as you like.

jquense:
o ok, you found the only inline styles :P yeah, just a consequence of doing the animation in Javascript.

I will take a look at the PR when I get a free moment, thanks for the contribution in any case.

PR #340, opened 2016-03-10
Add filter property to DateTimePicker

DiscoStarslayer:
For one of my projects I needed to restrict what characters were allowed in the DateTimePicker input field. This PR allows this functionality through a prop called filter.


jquense:
sorry I didn't get to this sooner. Contribution is/was appreciated!

PR #353, opened 2016-04-08
Messages can be nodes

taion:


jquense:
thanks

taion:
(It's just easier to use react-intl `<FormattedMessage>`s than to use the imperative API)

jquense:
gotcha, ya seems like a good use case

Issue #360, opened 2016-04-21
Cannot read property 'apply' of undefined


shapkarin:
Can I use browserify to load component? 

```javascript
var ReactWidgets = require('react-widgets/dist/react-widgets.js')
var NumberPicker = ReactWidgets.NumberPicker;
```

app.js:4466 Uncaught TypeError in console:
```javascript
app.js:4466 Uncaught TypeError: Cannot read property 'apply' of undefined
```

jquense:
yes but don't require them from `/dist` that is a global build for using in a script tag. import from `react-widgets` or `'react-widgets/lib/NumberPicker`

shapkarin:
the same error...
```javascript
var NumberPicker = require('react-widgets/lib/NumberPicker')
```

jquense:
I don't think the error is on the ReactWidgets side, something may be messed up with your build config. If you do think its a bug, please provide more info so I can reproduce it thanks!

shapkarin:
Ok. 
I use grunt to compile or 'npm start' and see the same error. 

grunt:
```javascript
browserify: {
    dist: {
      files: {
        'output/build/app.js': 'jsx/app.jsx'
      },
      options: {
        transform: ['reactify']
      }
    }
  }
```

npm:
```json
"scripts": {
    "start": "watchify jsx/app.jsx -v -t [ babelify --presets [ react ] ] -o output/build/app.js",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
```

Please help.

jquense:
I don't see anything strikingly odd, but I also don't ever use browserify, sorry there isn't much I can do to help.

foxhound87:
I got the same issue with version 3.4.4 using webpack.

`Uncaught TypeError: Cannot read property 'apply' of undefined`

`decimalChar @ localizers.js?9bf2:127`

```javascript
return (_numberLocalizer5 = _numberLocalizer).decimalChar.apply(_numberLocalizer5, arguments);
```

Grimfri:
Some error
`Uncaught TypeError: Cannot read property 'apply' of undefined`
`localizers.js:127`

nik-kor:
create a new issue https://github.com/jquense/react-widgets/issues/504

PR #373, opened 2016-04-28
Fix misleading propTypes error when no localizer configured

joshkel:
react-widgets is designed to throw an error if you try to use certain functionality when no localizer is configured. However, it didn't check "propType," so the propType validation function was undefined, resulting in React's propTypes validator throwing confusing messages like

> Warning: Failed propType: Cannot read property 'apply' of undefined Check the render method of `Uncontrolled(NumberPicker)`.

I believe this is the same problem seen in #360.


jquense:
thanks!

PR #376, opened 2016-04-29
Preserve text selection on tab

### LINKED ISSUE #375: Tabbing into formatted NumberPicker loses selection
Standard web browser behavior is to select all text within a text input element when tabbing to that element.  NumberPicker behaves this way by default; _however_, if a format is set, so the string value that's displayed while the input has the focus is different than the formatted value that's displayed normally, then switching from the formatted value to the display value loses the text selection.

joshkel:
Standard web browser behavior is to select all text within a text input element when tabbing to that element.  NumberPicker behaves this way by default; _however_, if a format is set, so the string value that's displayed while the input has the focus is different than the formatted value that's displayed normally, then switching from the formatted value to the display value loses the text selection.

The only idea I had for a solution was to check if all text is selected before update and, if that's the case, restore the selection after update.

Fixes #375.

### DIFF
--- a/src/NumberInput.jsx
+++ b/src/NumberInput.jsx
@@ -1,8 +1,15 @@
 import React from 'react';
+import ReactDOM from 'react-dom';
 import CustomPropTypes from './util/propTypes';
 import { number as numberLocalizer } from './util/localizers';
 
 let getFormat = props => numberLocalizer.getFormat('default', props.format)
+
+// http://stackoverflow.com/a/32598826/25507
+const canUseDOM = !!(
+  (typeof window !== 'undefined' &&
+  window.document && window.document.createElement)
+);
 
 export default React.createClass({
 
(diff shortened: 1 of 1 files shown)


src/NumberInput.jsx
```diff
@@ -56,6 +63,23 @@ export default React.createClass({
       this.getDefaultState(nextProps))
   },
 
+  componentWillUpdate(nextProps, nextState) {
+    if (canUseDOM) {
+      // Check if we need to restore text selection after pending update.
+      let node = ReactDOM.findDOMNode(this.refs.input);
+      this._selectAll = document.activeElement === node
+        && this.state.stringValue !== nextState.stringValue
```
jquense: 👋  hey there. I like the idea of this but i'm a bit worry about trying to mimic the browser behavior here. One concern is that this will maintain the selection at times that aren't the user tabbing into field, such as pasting.

joshkel: Sorry for the delayed reply.

I did some more testing, and I think it should be safe, with one exception. For the case of pasting, for example, the paste updates the field's contents and changes selectionStart, selectionEnd to the end of the text, then the onChange event fires, then componentWillUpdate and componentDidUpdate fire. So, by the time componentWillUpdate and compomentDidUpdate are fired, the selection is already at the end of the text, so the _selectAll behavior isn't triggered.

The only exception I've been able to find is if you trigger the spin button behavior by pressing Up or Down. In that case, the prop or state is changed directly, so all text is still selected when componentDidUpdate is fired, and the _selectAll behavior is triggered. Having the spin buttons preserve selection isn't necessarily bad, but it still feels a little odd, and it's inconsistent with the behavior of the spin buttons when some but not all text is selected.

The easiest solution, I think, would be to have NumberPicker.spin clear any selection before calling NumberPicker.change. I can add this logic if you're still open to this PR.

I posted some demo code I was using to test all of this to https://github.com/joshkel/react-widgets-tab.

PR #377, opened 2016-04-29
Expose select method

joshkel:
Exposing a select() method goes well with the existing focus() method and a standard <input> control's behavior.  See https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select.

It's possible for client code to work around the lack of a select()
method by doing

    ReactDOM.findDOMNode(this.refs.picker.refs.inner.refs.input).select();

but that's ugly and brittle.

### DIFF
--- a/src/NumberPicker.jsx
+++ b/src/NumberPicker.jsx
@@ -227,6 +227,10 @@ let NumberPicker = React.createClass({
     compat.findDOMNode(this.refs.input).focus()
   },
 
+  select() {
+    compat.findDOMNode(this.refs.input).select()
+  },
+
   increment() {
     return this.step(this.props.step)
   },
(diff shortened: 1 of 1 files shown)


jquense:
hey! thanks a lot for taking the time to put this together! I don't know that exposing more imperative DOM api's is a great idea...in general its not idiomatic, though sometimes necessary. I'm more prone to say that using a findDOMNode and queryselector is a better option. It's certainly ugly and more work, but I think direct DOM manipulation probably should be a bit painful since you're off the happy path.

I do realize that exposing the `focus()` is a bit hypocritical but in that case its a _very_ common action that has no good declarative api. `select` on the other hand is a more niche method.

joshkel:
Fair enough. I'll stick with `findDOMNode` + `querySelector`. Thanks!

PR #386, opened 2016-05-12
Guard against this.refs.input being null

amoenk:


jquense:
thanks for the PR, it looks like you actually made the change in the wrong place, `/lib` is compiled code, please make the change in `/src` so it doesn't get overridden next time we build!

PR #392, opened 2016-05-31
added dependency loose-envify

6ewis:
Also, note that loose-envify has a envify as a dependency so we don't need to include it


jquense:
thanks!

PR #393, opened 2016-06-07
Fix case insensitive add term bug

### LINKED ISSUE #279: When "caseSensitive" is false, adding a term should not be suggested
www

oyvinmar:
Fixes #279


jquense:
thanks!

PR #396, opened 2016-06-09
Remove Duplicated Class from Combobox's input

brock-noah:
`Combobox` passes `rw-input` to the `ComboboxInput` input as a className.
`ComboboxInput` appends `rw-input` (`" rw-input"`) to that className prop. 
Causing a result of `<input class="rw-input rw-input" ... />`.

Future Topic:
- import cx in ComboboxInput to remove string concatenation


jquense:
thanks!

Issue #408, opened 2016-07-01
React 15.2.0 and unknown props


tylergoodman:
Several of the widgets pass down invalid props as described [here](https://fb.me/react-unknown-prop). Here's some of the ones I've run into:

In Multiselect:
> Warning: Unknown props `caseSensitive`, `ariaActiveDescendantKey` on `<div>` tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
>     in div (created by Multiselect)
>     in Multiselect (created by Uncontrolled(Multiselect))
>     in Uncontrolled(Multiselect)

> Warning: Unknown prop `searchTerm` on `<input>` tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
>     in input (created by MultiselectInput)
>     in MultiselectInput (created by Multiselect)
>     in div (created by Multiselect)
>     in div (created by Multiselect)
>     in Multiselect (created by Uncontrolled(Multiselect))
>     in Uncontrolled(Multiselect)

> Warning: Unknown props `onOpening`, `duration`, `onClosing`, `onClose`, `onOpen` on `<div>` tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
>     in div (created by Popup)
>     in Popup (created by Multiselect)
>     in div (created by Multiselect)
>     in Multiselect (created by Uncontrolled(Multiselect))
>     in Uncontrolled(Multiselect)

and in DateTimePicker:

> Warning: Unknown props `editFormat`, `editing`, `culture`, `parse`, `textValue` on `<input>` tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
>     in input (created by DatePickerInput)
>     in DatePickerInput (created by DateTimePicker)
>     in div (created by DateTimePicker)
>     in DateTimePicker (created by Uncontrolled(DateTimePicker))
>     in Uncontrolled(DateTimePicker)

vdh:
Also triggers for `busy` prop on `Multiselect`

vdh:
It looks like this was fixed in v3.3.4 by 62c1fe6

FBosler:
This isn't fixed for me. How can I fix it?

> Warning: Unknown props `editFormat`, `editing`, `culture`, `parse`, `textValue` on <input> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
>     in input (created by DatePickerInput)

[...]

PR #409, opened 2016-07-06
Added rw-img-path variable

vdh:
Similar to the existing `rw-font-path` variable, it would help to be able to adjust the paths to the loading images, especially when the default relative paths don't resolve correctly (for example, when loaded via `sass-loader` in Webpack).


vdh:
The test failure looks to be caused by #408.

jquense:
LGTM thanks

PR #411, opened 2016-07-07
Remove unnecessary tilde escape strings

vdh:
Removed unnecessary tilde escape strings in `icons.less` for the `url`s in the `@font-face` definition, and one in `core.less` for one of the box-shadow insets. The last escape in `core.less` at [L285](https://github.com/vdh/react-widgets/blob/d46b9b0eefa4114437d7da9ee0cf3e88de1a7526/src/less/core.less#L285) still looks to be necessary since the build stalls without the escape, presumably because of the comma in the value.

### DIFF
--- a/src/less/core.less
+++ b/src/less/core.less
@@ -353,7 +353,7 @@ ul.rw-list {
 .rw-combobox input.rw-input,
 .rw-datetimepicker input.rw-input,
 .rw-numberpicker input.rw-input {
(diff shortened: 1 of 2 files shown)


vdh:
Rebased to the latest master branch, and remove the changes to `lib/*` and `dist/*` files.

jquense:
thanks!

PR #417, opened 2016-07-11
Update icons.scss

saxicek:
url() in double quotes does not work. Probably the script less2scss does not transform `src: ~"url('@{rw-font-path}/rw-widgets.eot?v=@{rw-version}')";` properly. Unfortunately I don't know less syntax to suggest a fix.

### DIFF
--- a/lib/scss/icons.scss
+++ b/lib/scss/icons.scss
@@ -1,11 +1,11 @@
 
 @font-face {
   font-family: 'RwWidgets';
-  src: "url('#{$rw-font-path}/rw-widgets.eot?v=#{$rw-version}')";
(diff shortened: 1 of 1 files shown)


vdh:
I've fixed the LESS code that generates this in #411

jquense:
👍

saxicek:
I guess the SCSS needs rebuild, because the error I was fixing is still present in current release. Do you plan to release a new version with this issue fixed? Thanks!

vdh:
@saxicek I made temporary [`vdh/react-widgets#v3.4.1-temp-fork`](https://github.com/vdh/react-widgets/tree/v3.4.1-temp-fork) branch with a rebuild in it while I waited for a new official one.

PR #430, opened 2016-07-26
Improve keyboard navigation performance for open dropdown

joshkel:
Once a DropdownList has a few hundred items or so, using the keyboard to
navigate an open dropdown list gets _extremely_ slow.  Adding
shouldComponentUpdate dramatically improves performance.

Rather than trying to add shouldComponentUpdate to ListOption (which
would require that ListOption know how its callers are using it and what
its children are), this PR refactors some duplicate code out of List and
ListGroupable into a new ListOptionItem component and adds update
checking there.

### DIFF
--- a/src/List.jsx
+++ b/src/List.jsx
@@ -1,10 +1,10 @@
 import React   from 'react';
 import ListOption from './ListOption';
+import ListOptionItem from './ListOptionItem';
 import CustomPropTypes from './util/propTypes';
 import compat from './util/compat';
 import cn from 'classnames';
 import _  from './util/_';
-import { dataText, dataValue } from './util/dataHelpers';
 import { instanceId, notify } from './util/widgetHelpers';
 import { isDisabledItem, isReadOnlyItem }  from './util/interaction';
 
@@ -90,7 +90,7 @@ export default React.createClass({
             , isReadOnly = isReadOnlyItem(item, this.props);
 
           return (
-            <Option
+            <ListOptionItem
               key={'item_' + idx}
               id={currentId}
               dataItem={item}
@@ -98,19 +98,12 @@ export default React.createClass({
               readOnly={isReadOnly}
               focused={focused === item}
               selected={selected === item}
-              onClick={isDisabled || isReadOnly ? undefined : onSelect.bind(null, item)}
-            >
-              { ItemComponent
-                ? <ItemComponent
-                    item={item}
-                    value={dataValue(item, valueField)}
-                    text={dataText(item, textField)}
-                    disabled={isDisabled}
-                    readOnly={isReadOnly}
-                  />
-                : dataText(item, textField)
-              }
-            </Option>
+              onSelect={onSelect}
+              optionComponent={Option}
+              itemComponent={ItemComponent}
+              textField={textField}
+              valueField={valueField}
+            />
           )
         });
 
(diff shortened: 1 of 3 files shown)


jquense:
Hey there, thanks for the PR, performance improvements are always appreciated.  There is a bit of a problem however.  this will break SelectList,  which wraps ListOption and passes in additional props. those changes wouldn't be flushed because the new outer Component will now block updates. 

This approach may work by refactoring ListOption and using shallow compare instead of explicit props checks but that may negate some of the improvements

joshkel:
Sorry. I should have noticed SelectList.

I'm having trouble thinking of a good way of accommodating this. I can think of a few approaches:

1. Move shouldComponentUpdate to ListOption, and do a shallow compare (so that it can handle additional props, like SelectList's). The problem is that the ListOption's children are going to be different each time it's rendered (if I understand React correctly), so a simple shallow compare would be inadequate. It'd have to exclude `children` from the comparison, and it would have to receive `textField`, `valueField`, and `itemComponent` itself, just so that it could detect updates to its child. All of that seems ugly.
2. I could stick with the ListOptionItem approach in this PR and change shouldComponentUpdate to always return true if it's using a non-default listComponent. (I.e., if it's not the default/known/standard component, it doesn't even try to optimize.) This would mean that SelectList doesn't benefit from the optimization, but I expect that SelectLists aren't going to be used for large numbers of items, so I don't think that should be a problem.
3. I could add SelectList's `checked` and `role` props to the base `List` and `ListGroupable` components and the base ListOption component, so that SelectList's dynamically created kind-of-impure list item class doesn't need to do nearly as much, so a simple ListOptionItem approach would work. This clutters the List/ListGroupable/ListOption design a bit, but it seems reasonable (not excessively complicated).

The second approach would probably be my pick, because it's simple and self-contained and doesn't ripple out to the rest of the design. Would you be open to one of these approaches? Or is there another approach that I'm overlooking? I'll update my PR with whatever approach you prefer.

ZhangScott:
Hi Jason. 

Sorry for bothering you. It is not a right place to ask you question about 'react-widget', but I can't contact you via mail. 

I am developing the 3rd party web control, and using the 'react-widget'. Currently I would like to customize my control via overwriting the less variables, other than overwriting css directly, it makes more succinct in code.  My question is that is it possible to import the 'react-widget' less file directly and re-value a variable, then it take effect on the webpage control without any css compiling?

Thanks,
Yuxiong

ZhangScott:
Additionally, I have tried @import '../../../node_modules/react-widgets/lib/less/bootstrap-theme.less';  and set the variable value, but it does not take effect. Did I miss some important steps?

jquense:
TBH, when it comes to long lists in these widgets I think the more robust solution is virtual-lists, using something like react-virtualized. I don't provide those out of the box but the current API is designed to allow that sort of customization (i.e. i've built those lists at work)

I do think that option 2 is probably fine, given that ListOption is a really low level API to be messing with, and the author can add the SCU to their own component if needed. I would really like to avoid adding _another_ component though but that seems not very feasible, at least we should name it something  that indicates it only exists to add an optimization to the list items.

jquense:
hmm it might be worth holding off a bit on this. I may be rewriting how Lists work in a big way soon. The current contract is messy.

PR #448, opened 2016-08-28
fix for manual proptypes call on finalView

SorenPeterson:
In a future release of React, manually calling PropTypes will not be allowed. Currently it is generating a warning. There is a secret argument that they are using to detect manually calling PropTypes. This pull request implements that fix.

Source: https://facebook.github.io/react/warnings/dont-call-proptypes.html


jquense:
thanks for hte PR, you've actually changed the compiled, generated code here. this change will get overridden next time the code is built. you want to make your changes in `src/`

PR #453, opened 2016-09-02
Add ability to use a function as the step prop in NumberPicker

Binck360:


Binck360:
I'll create a separate branch for this and open a new pull request

PR #454, opened 2016-09-02
Document false value of onCreate.

hjnilsson:


jquense:
thanks for the PR, there isn't really a need to set hte value to `false` any falsely value will have this effect. INn other words the logic is "only allow new tag creation when an `onCreate` is provided`

PR #458, opened 2016-09-07
Mouse wheel support in number picker

Binck360:
This adds mousewheel support to the numberpicker. Scrolling up increments the value, scrolling down decrements it.

### DIFF
--- a/dist/react-widgets.js
+++ b/dist/react-widgets.js
@@ -88,7 +88,6 @@ return /******/ (function(modules) { // webpackBootstrap
 /***/ function(module, exports) {
 
 	// shim for using process in browser
-
 	var process = module.exports = {};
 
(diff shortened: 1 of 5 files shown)


jquense:
thanks for the PR! I see why you might want this, however I don't think it's a good fit for the core library. overloading the scroll wheel with another function besides scrolling can be confusing, since it breaks user expectations if you want to start scrolling while the mouse is over an input.

Binck360:
Hi, is there a way to implement this in another way that you would be interested in accepting in the core lib? Our app is always fullscreen so it's not that big of an issue. Or is there a way to implement this without changing anything to the core lib maybe?

jquense:
you can wrap the NumberPicker and add your own onWheel handler to increment/decrement the value. Essentially just use the same logic here but in a wrapping component

PR #462, opened 2016-09-15
Add support for more semantically correct icons

kjschulz:
The `Button` and `Multiselect` components utilize the popular `<i>` tag for icons. If a project is required to pass 508 compliance testing before shipping to production, all components inheriting either of the above mentioned components as a base will fail. This PR causes 508 compliance tests to pass by taking a little advice from Font Awesome as referenced here: http://fontawesome.io/examples 

> ...we like the `<i>` tag for brevity, but using a `<span>` is more semantically correct.

### DIFF
--- a/dist/react-widgets.js
+++ b/dist/react-widgets.js
@@ -1942,7 +1942,7 @@ return /******/ (function(modules) { // webpackBootstrap
 	        'aria-label': label,
 	        className: (0, _classnames2.default)(className, 'rw-btn', active && !disabled && 'rw-state-active')
 	      }),
-	      icon && _react2.default.createElement('i', {
+	      icon && _react2.default.createElement('span', {
 	        'aria-hidden': true,
 	        className: (0, _classnames2.default)('rw-i', 'rw-i-' + icon, busy && 'rw-loading')
 	      }),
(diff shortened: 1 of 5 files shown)


jquense:
can you talk about why they fail? semantically correct here is sort of meaningless for presentational icons. and any a11y concerns are handled by labels and roles

kjschulz:
While I don't necessarily agree with the rule, using the `<i>` tag for presentational icons fails under this category:

> Semantic markup should be used to mark emphasised or special text so that it can be programmatically determined. - `Section508.D.[msgCode]`

Although it's generally agreed to be outdated in practice, it seems that the trouble arises from the fact that the `<i>` tag was originally created for creating emphasis to a portion of text. Equivalent in wording to WCAG 1.0 checkpoint 6.1, the primary intention appears to be to accommodate users that require a default stylesheet to be turned off, to read the raw document, or to replace it with their own stylesheet.

kjschulz:
@jquense any thoughts about merging this in?

jquense:
thanks

vdh:
@jquense @kjschulz This commit has broken the Multiselect busy styles, because the LESS file wasn't updated (still targets `i.rw-loading`)

kjschulz:
@vdh @jquense What is the correct class?  The documentation page still appears to be using `rw-loading`?  If you tell me what it should be, I'll be happy to patch and PR the change.

jquense:
I think it's jsut that the selector includes the tag name

kjschulz:
Gotcha, fixing now...my bad.

kjschulz:
@jquense @vdh  Here's the PR to fix:  https://github.com/jquense/react-widgets/pull/543

PR #468, opened 2016-09-25
Change watching attribute from key to keyCode for Japanese IME in Multiselect

goro56:
Japanese user's usually input text with IME (Input Method Editor), and decide word by Enter key. But Multiselect can't distinguish enter for IME and decide tag of Multiselect.

so Multiselect should watch not key attribute but keyCode attribute. because when decide word in IME, keyCode attribute will be "229" but in other cases it will be "13".


jquense:
thanks for the contribution. Looks good to me, can you add a comment next to the line explaining why it uses the keycode instead of the key?

goro56:
thanks for your understanding. I added a comment.

jquense:
Thanks!

PR #480, opened 2016-10-07
Allow setting default time

michaeljacobdavis:
It looks like you can currently set a "default start date" for calendar, but not time. The structure is already in place to allow:

```
<DateTimePicker
  currentDate={new Date('1/1/1970 8:00 AM')}
  calendar={false} />
```

to open the time picker at 8 AM:

This just adds a fallback to `currentDate` for `focusedItem` when `value` doesn't exist.


jquense:
thanks!

Issue #473, opened 2016-10-02
Date picker causing safari to lock up


ryanking1809:
The react widget date picker is locking up my Safari Version 10.0 (12602.1.50.0.10)
It seems to work fine on browser stack but we're hearing similar issues from other users.

There's no errors in the console indicating what going on. Cleared caches, removed extensions, and it still locks up (all you see is the spinning wheel of death & have reload the tab to get response).

Is anyone else having issues / have a solution to what going on?

jquense:
working for me as well. I haven't heard anything either :/

jheander:
We have plenty of similar reports from users running both Safari and Internet Explorer. I have managed to reproduce the problem by putting a date picker in an area that uses either "position: fixed" or "position: absolute".

This causes the code to get stuck in a loop when you try to open the calendar in the date picker. The loop keeps coming back to these lines in Popup.js:126

```
var height = this.height();
if (height !== this.state.height) this.setState({ height: height });
```

I added log prints here, and it seems that the calculated height and the current state is oscillating between to very similar values, differing only at the third decimal. But since they are compared to absolute equality it will go on for ever. Such is the peril of comparing floating point numbers :)

I'll send a pull request that fixes the issue. Maybe it is the same issue you are seeing ryanking1809.

jheander:
@jquense Would it be possible to do a NPM-release of this 🙏 ? Currently running all projects on a forked repo to avoid lockups in safari and IE.

jquense:
should be released now

PR #488, opened 2016-10-24
Safer floating point comparison to avoid infinite update loop

jheander:
Solves #473 by avoiding potential infinite loop for browsers and conditions where the height calculation does not give a stable exact floating point value.

### DIFF
--- a/lib/Popup.js
+++ b/lib/Popup.js
@@ -122,21 +122,9 @@ exports.default = _react2.default.createClass({
       return;
     }
 
-    if (opening) {
-      this.open();
-    } else if (closing) {
-      this.close();
-    } else if (open) {
-      // this.height() returns a floating point number with the desired height
-      // for this popup. Because of potential rounding errors in floating point
-      // aritmetic we must allow an error margin when comparing to the current
-      // state, otherwise we can end up in an infinite loop where the height
-      // is never exactly equal to our target value.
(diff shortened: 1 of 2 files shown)


lib/Popup.js
```diff
@@ -124,7 +124,10 @@ exports.default = _react2.default.createClass({
 
     if (opening) this.open();else if (closing) this.close();else if (open) {
       var height = this.height();
-      if (height !== this.state.height) this.setState({ height: height });
+      var diff = Math.abs(height - this.state.height);
+      if (isNaN(diff) || diff > 0.1) {
```
jquense: can you add a comment describing what this is fixing so it doesn't get removed later accidentally?

jheander: Sure! I have added a comment and structured the surrounding if-statement a bit since this block now became pretty long. You could also handle the problem by rounding the target size returned by this.height() to the nearest integer, but I guess browsers actually have sub-pixel positioning so we might want the exact value?

jquense:
sorry just noticed. You edited teh wrong file, you want to edit the files in `src/` not `lib/` lib is the generated compiled output

jheander:
Ah, I thought the code formatting was a bit strange :) Pushed new patch!

jquense:
thanks!

agile116:
@jquense Can we release a new version to npm? We encountered the same issue in IE. Thanks.

PR #490, opened 2016-10-25
Update interaction.js

cyrusbowman:
If `disabled`/`readOnly` is not an array don't assume we want to disable the matching value. In the case of DropdownList any value supplied to `disabled` that is not an array, true, false, or null, throws an invalid prop warning message anyways.

### DIFF
--- a/src/util/interaction.js
+++ b/src/util/interaction.js
@@ -21,7 +21,7 @@ export function isReadOnlyItem(item, props){
 export function contains(item, values, valueField){
   return Array.isArray(values)
       ? values.some(value => valueMatcher(item, value, valueField))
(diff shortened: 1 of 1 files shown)


src/util/interaction.js
```diff
@@ -21,7 +21,7 @@ export function isReadOnlyItem(item, props){
 export function contains(item, values, valueField){
   return Array.isArray(values)
       ? values.some(value => valueMatcher(item, value, valueField))
-      : valueMatcher(item, values, valueField)
+      : false
```
jquense: this behavior is needed elsewhere, this should be managed in the isDisabled/isReadOnly helpers above

jquense:
closing for now, feel free to update

PR #491, opened 2016-10-27
Fixed active style issue for DateTimePicker's buttons

vdh:
I noticed when both date and time buttons are available on the `DateTimePicker` component, the active styles will show both buttons pressed when one is pressed.

It reminded me of an earlier issue with `NumberPicker`'s button styles (#442), so I re-implemented the same active state logic that currently exists in `NumberPicker` into `DateTimePicker` so that the active state of the buttons could be separated in a similar way.


jquense:
loooong delay, sorry about that :P I think the current logic is ok for now

PR #495, opened 2016-11-01
FIXED #493 - Add property to Combobox to be able to sort the items based on the search term

### LINKED ISSUE #493: Add property to Combobox to be able to sort the items based on the search term
Add a property to the Combobox widget that allows either a boolean or a function.

When `true` is supplied as value, the items will be sorted based on their data text value.

When a function is supplied as value, the function is used as comparison function with the arguments: `itemA, itemB, searchTerm`

arnogeurts:
Fixes #493

### DIFF
--- a/dist/react-widgets-simple-number.js
+++ b/dist/react-widgets-simple-number.js
@@ -88,11 +88,9 @@
 	};
 
 	function simpleNumber(options) {
-	  var _defaults$options = _extends({}, defaults, options);
-
-	  var decimal = _defaults$options.decimal;
-	  var grouping = _defaults$options.grouping;
-
+	  var _defaults$options = _extends({}, defaults, options),
+	      decimal = _defaults$options.decimal,
+	      grouping = _defaults$options.grouping;
 
 	  var localizer = {
 	    formats: {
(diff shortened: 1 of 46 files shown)


jquense:
thanks for the contribution! ultimately I think this sort of thing is best left out of the core library here. It's simple enough to implement without changes here with a wrapping component and controlling `searchTerm` unless I'm misunderstanding the feature request (let me know}

arnogeurts:
Thanks for your response. 

I agree that it would be possible to do this with a wrapping component, however, we are using the react widgets in some of our projects, and multiple of our customers have requested this feature. 

Let's for example say that we have a combobox containing a list of the states of the United States of America in alphabetical order and has a filter function that filters based on a substring match. When I type in "Or" as search term, it is clear that I am trying to find "Oregon", however "Colorado" would be my first match. In this case it would be very useful to be able to sort the data in the combobox based on the most likely values.

jquense:
I appreciate the contribution here, and the your use case! I still think that this is best handled outside the library, since its 1. pretty straight forward encapsulate the logic in your own component (wrapping this one), 2. its hard to pick good defaults for "most likely" that makes sense across languages, and cultures and use cases.

Thanks again though for taking the time to put this together!

PR #506, opened 2016-12-01
Use byteLength instead of length for MultiselectInputt size

yosuke-furukawa:
When multi-byte user inputs this MultiselectInput form, Our character has doubled size to compare with ascii characters. We need to use `byteLength` instead of `length`.

### DIFF
--- a/src/MultiselectInput.jsx
+++ b/src/MultiselectInput.jsx
@@ -17,7 +17,7 @@ class MultiselectInput extends React.Component {
 
   render() {
       let { disabled, readOnly, ...props } = this.props
-      let size = Math.max((props.value || props.placeholder).length, 1) + 1;
+      let size = Math.max((props.value || props.placeholder).byteLength, 1) + 2;
 
       return (
         <input


jquense:
hey there, thanks for the PR. Looks like the change you've made here doesn't work. I don't think that `byteLength` is a property on strings, it isn't on Chrome.

Also I'm sure that if it did that it'd be the correct thing. the input size attribute is "number of characters" not bytes

yosuke-furukawa:
oooops, sorry for that.
As I said, this `size` property is not suitable with multi-byte characters.
So I would like to add an option to disable this `size` attribute, like `ignoreSize` props.
If you are comfortable to add an option, I will send another PR.

jquense:
the problem is that if you ignore the size the text input won't grow, ever.

yosuke-furukawa:
I've got it, However, input size default is 20. we don't need to input the size > 20. 
However, this size check is not fixed the problem so I close this.

PR #511, opened 2016-12-12
Select items render outside of parent when scrolling 

Flamenco:
Please see screen shots for example of items still showing outside of parent container when scrolling.

screen shot 2016-12-12 at 11 25 38 am

screen shot 2016-12-12 at 11 25 55 am

### DIFF
--- a/lib/scss/selectlist.scss
+++ b/lib/scss/selectlist.scss
@@ -7,6 +7,7 @@
   @extend ul.rw-list;
 
   > ul {
+    position: relative;
     height: 100%;
     overflow: auto;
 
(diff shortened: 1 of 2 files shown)


jquense:
thanks for the PR! I have to look at this a bit since I've never seen the behavior your seeing

Flamenco:
It's nested pretty deep in my use case, but somehow a parent within the widget hierarcy does not have absolute or relative positioning set.

This is on chrome BTW

Flamenco:
Here's another issue with the relative position propagating up to a parent beyond the react component.

screen shot 2016-12-14 at 5 35 11 pm

screen shot 2016-12-14 at 5 35 33 pm

PR #517, opened 2016-12-22
Get size of MultiselectInput by multibyte-length

gorohash:
this PR is another solution for #506 .

currently

this PR

we can't see all of input letters, if we input some multi-byte letters in Multiselect component.
so we need to regard a multi-byte letter as 2 letters, when we count size of input.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "dom-helpers": "^2.2.4",
     "invariant": "^2.1.0",
     "loose-envify": "^1.2.0",
+    "multibyte-length": "^0.1.1",
     "uncontrollable": "^4.0.0",
     "warning": "^2.0.0"
   },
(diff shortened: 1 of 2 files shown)


jquense:
sorry for the delay, I still am not sure that this is the right approach. "byte length" doesn't correspond to "text width" there are many multi-byte characters that are not wider then a normal character.

gorohash:
thank you for your response.

I agree with you in principle. but I think this approach is better than others that I could think of.
I would like to hear your opinion for the following ideas.

* replace `input` with content editable `span`
* add an option to set `size` attribute

jquense:
yeah it's not an easy problem :/ 

I can't really speak to whether content editable is a good idea at the moment (i'd have to explore it) But I do think that making `size` configurable is fine for now, maybe a function that takes the current string and returns a size?

gorohash:
thank you for your advice!

I'll send another PR. so I close this one.

PR #518, opened 2016-12-29
Add inputSize option to Multiselect

gorohash:
this PR is another solution for #517 , #506 .

I added `inputSize` option to Multiselect for set size of MultiselectInput.
`inputSize` is a function that takes current string and return `size` of `input`.

if this is good to you, I'll add docs for this.


jquense:
thanks!

Issue #472, opened 2016-09-30
itemComponent aka timeComponent not working


AgamlaRage:
Hi,

I use **react-widgets@3.4.4** and I notice that the options **timeComponent** aka **itemComponent**, is not working.
I can't set a custom format to my timelist items like in your demo.

[http://jquense.github.io/react-widgets/docs/#/datetime-picker?_k=pberim](http://jquense.github.io/react-widgets/docs/#/datetime-picker?_k=pberim)
![sans titre](https://cloud.githubusercontent.com/assets/15155827/19009437/400376bc-8774-11e6-96e2-05e2b28055a1.png)

To be really sure, I downloaded the react-widget git to try this demo in local.
Same result, I was unable to use this options.

There are missing code in **src/TimeList.jsx**
The declaration of **itemComponent** is well made, but the props is not passed to the child component **<List/>**

I'm not familiar with push/pull request, then sorry for that.

**Here is the changes I made in the demo to make it works :** src/TimeList.jsx
- **line74** `let { value, onSelect, itemComponent } = this.props; // add itemComponent`
- **line80** `itemComponent={itemComponent} // add new props to <List/> declaration`

**Here is the changes I made in my app to make it works :** node_modules/react-widgets/lib/TimeList.js
- **line94** `var itemComponent = _props.itemComponent; // add itemComponent`
- **line101** `itemComponent: itemComponent // add new props to <List/> declaration`

brandonfancher:
I second this.

PR #521, opened 2017-01-06
Fix timeComponent prop for customizing time dropdown for DateTimePicker.

brandonfancher:
This fixes the problem referenced in Issue #472. I've only done a couple OSS PRs, so feel free to let me know if I can do anything better next time.


jquense:
thanks!

PR #524, opened 2017-01-16
Update moment localizer to use strict mode

### LINKED ISSUE #523: DatePicker / Moment localization - doesn't use strict mode
The third parameter to the moment constructor is a boolean signifying strict mode.

When not present, a lot of inconsistent date parsing occurs when there is a difference between the view format and edit format.  The parser attempts both formats, first view then edit.  If one of them returns `null` will will try the next.  Moment outside of strict mode will try its darndest to return *something* even if it's not entirely accurate.  Consider the following:

```jsx
<DateTimePicker
  format="MMM D YYYY"
  editFormat="DD/MM/YYYY" 
/>
```

When entering "01/01/2017" in the picker, which is the expected edit mode format - the date of "Jan 1 2001" will be returned.  Wait, what?  This is because moment is not in strict mode... for better or worse,  

```js
moment("01/01/2017", "MMM D YYYY").getDate().toJSON() // => 2001-01-01T08:00:00.000Z  (that's PST)
```

If in the `getMoment` function, if `true` is passed as the third parameter to the moment constructor and the date is not in the expected format, it will return `null` if , similar to globalize.

tswaters:
Fixes #523 

A few things

* I didn't see any guidelines on committing to ./dist and lib, so I went ahead and ran the build and included relevant changes in dist/lib.

* When I ran it I saw some changes that were unrelated (weird), I left those out e.g.:

```diff
-       var dates = _extends(_dateArithmetic2.default, {
+       var dates = _extends({}, _dateArithmetic2.default, {
```

They all seemed to tie to using extends.

* Tests appear to pass

* I've also submitted the same PR to react-widgets-moment-localizer, see:

https://github.com/jquense/react-widgets-moment-localizer/issues/6
https://github.com/jquense/react-widgets-moment-localizer/pull/7

### DIFF
--- a/dist/react-widgets-moment.js
+++ b/dist/react-widgets-moment.js
@@ -73,7 +73,7 @@
 	  if (!hasLocaleData) throw new TypeError('The Moment localizer depends on the `localeData` api, please provide a moment object v2.2.0 or higher');
 
 	  function getMoment(culture, value, format) {
-	    return culture ? moment(value, format)[localField](culture) : moment(value, format);
+	    return culture ? moment(value, format, true)[localField](culture) : moment(value, format, true);
 	  }
 
 	  function endOfDecade(date) {
--- a/lib/localizers/moment.js
+++ b/lib/localizers/moment.js
@@ -11,7 +11,7 @@ exports.default = function (moment) {
   if (!hasLocaleData) throw new TypeError('The Moment localizer depends on the `localeData` api, please provide a moment object v2.2.0 or higher');
 
   function getMoment(culture, value, format) {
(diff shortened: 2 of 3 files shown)


jquense:
hey there 👋  I think this is a good change, but also a breaking one, and given how the localizers are included with the main widget package it means I can't include it in v3. That said tho, v4 is almost out and has separated out the packages a bit to avoid these issues in the future.

All that being said, can you make this change against `v4` branch instead of master?

tswaters:
Ok, will do.  I'll close this one and open another built against v4

PR #525, opened 2017-01-17
Update moment localizer to use strict mode

### LINKED ISSUE #523: DatePicker / Moment localization - doesn't use strict mode

tswaters:
Fixes #523 

As requested - I've based this off v4 branch.
Again, updated dist directory as part of the commit.

Cheers,


jquense:
thanks!

PR #531, opened 2017-01-31
Fix webpack sass error by removing root-level child selector

jmarks1992:
Currently, when building with webpack we get the following build error:

```
WebpackError: 
            ModuleBuildError in /node_modules/css-loader/index.js!/node_modules/sass-loader/index.js!/node_modules/react-widgets/lib/scss/react-widgets.scss
            Module build failed: 
&.rw-state-disabled {
^
      Base-level rules cannot contain the parent-selector-referencing character '&'.

Backtrace:
	node_modules/react-widgets/lib/scss/core.scss:139
      in /node_modules/react-widgets/lib/scss/core.scss (line 139, column 1)
```
By removing this character I was able to remove the compilation error.

### DIFF
--- a/lib/scss/core.scss
+++ b/lib/scss/core.scss
@@ -136,7 +136,7 @@
   @include state-select();
 }
 
-&.rw-state-disabled {
+.rw-state-disabled {
   @include state-disabled();
   opacity: 1;
 }
--- a/src/less/core.less
+++ b/src/less/core.less
@@ -136,7 +136,7 @@
   .state-select();
 }
 
-&.rw-state-disabled {
+.rw-state-disabled {
   .state-disabled();
   opacity: 1;
 }


pelzhaus:
lifesaver - THANK YOU

jquense:
thanks however this isn't gonna fix the issue long term. The SCSS files are generated from the LESS files in `src/less` we need to fix those files or the transpiltion script otherwise this will just reappear on the next build

jmarks1992:
@jquense the same &.rw-state-disabled is present in core.less under src/less

```
&.rw-state-disabled {
  .state-disabled();
  opacity: 1;
}
```

Could that rule just be modified there?

jquense:
o i didn't realize the nesting selector was in the root scope like that. I'm really surprised that LESS just was ok with that. This should be fine

vdh:
Is this going to see a release sometime soon? This is quite the thorn to builds.

PR #533, opened 2017-02-01
removed duplicate dependancies

zeroasterisk:
uncontrollable was in both prod and dev dependancies. new versions of npm will prune it, and it just kinda doesn't make sense.


jquense:
thanks!

Issue #176, opened 2015-08-01
Disable input on DateTimePicker


sporto:
I would like to disable the text input in the DateTimePicker component, still want the user to be able to click on the calendar and time pickers. Is there a setting for this I am missing?

jquense:
Hi there, no there isn't anything you are missing that particular combination is not directly supported but you can fake it by wrapping the Component and in `componentDidMount` manually disabling the input. It's not a great option but its the best I can think of.

sporto:
Thanks, I will try this. Please consider adding a props for this use case.

jquense:
I will keep it mind for the future thanks :)

PR #542, opened 2017-02-16
Add readOnlyInput option to DateTimePicker

facto:
I have a common use case where I want to tightly constrain the user's ability to specify a date and/or time. This option makes only the input element read-only. This makes it very easy to force the user to, e.g., stay within the time step limit.

Related: #176


jquense:
this was added in the most recent version via `inputProps`

PR #543, opened 2017-02-17
Multiselect busy styles fix

kjschulz:
Fixed broken Multiselect busy styles due to outdated html tag reference


vdh:
Sweet, thanks! I think there's also a `lib/less/multiselect.less` file generated from the `src/less/multiselect.less` one? I prefer Sass personally, but it's always good to be thorough :)

jquense:
anything in the lib directory is generated, so no need to change it there :)

kjschulz:
Ok, so do I need to remove ` lib/scss/multiselect.scss ` and just commit ` src/less/multiselect.less ` and `dist/css/react-widgets.css `?

PR #546, opened 2017-03-02
add htmlFor and id to selectList inputs

Racheldelman:
Hi,
can you merge this in?


jquense:
hi there. I don't think there is a need for this. The input being the child of the label is enough to associate them, you don't also need a `for` attribute

Racheldelman:
Ok thanks. I didn't realize that I could style it based on that

PR #556, opened 2017-04-03
Fixed ComboBox#value documentation's example

expobrain:


jquense:
thanks for the PR but i'm not sure what you are trying to fix here. The current example is correct

expobrain:
Actually the `Combobox` component doesn't have any `defaultValue` props defined, so the documentation says to use the `value` prop but the example shows the usage of `defaultValue`. This was confusing to me that's why I made this PR.

jquense:
yeah I can see why that's confusing, I'll see if I can clarify in the future on the docs. `defaultValue` is the "same" prop as `value` but for uncontrolled inputs. You can see that it's mentioned to the right (in pink) of the Value prop header. It works the same as defaultValue and value on normal react inputs. http://jquense.github.io/react-widgets/docs/#/controllables?_k=z00zlu

PR #580, opened 2017-05-26
Replaced React.createClass with create-react-class

joshgagnon:
Also have replaced isMounted with checks to this._mounted, which is managed by life cycle functions.

### DIFF
--- a/dist/react-widgets.js
+++ b/dist/react-widgets.js
@@ -8,7 +8,7 @@
 		exports["ReactWidgets"] = factory(require("react"), require("react-dom"));
 	else
 		root["ReactWidgets"] = factory(root["React"], root["ReactDOM"]);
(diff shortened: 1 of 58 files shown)


zerkms:
A note from the react documentation:

> Along with functional components, JavaScript classes are now the preferred way to create components in React.

What is the reason to not use the officially preferred way with ES2015 classes, but ancient createClasses thing?

joshgagnon:
The reason is this library uses a handful of mixins, and would need a major revision to move over to ES2015 classes.

jquense:
Thanks!

ericvilleneuve:
Any chance we can get a 3.4.9 release including this PR?

sylvain-hamel:
@jquense would you please release the npm package with this fix? Thanks

waynebrantley:
@jquense Any changes for npm package?   Or are you just going to release 4.0 instead?

Issue #559, opened 2017-04-10
deprecations in react 15.5.0 


tswaters:
See the following for details: 

https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

createClass has moved from `react` to another package --

```diff
-var Component = React.createClass({
+var createReactClass = require('create-react-class');
+var Component = createReactClass({
```

propTypes have moved from `react` to another package --

```diff
-import React, {PropTypes} from 'react'
+import React from 'react'
+import PropTypes from 'prop-types';
```

`react-addons-test-utils` has moved to `react-dom/test-utils`:

```diff
-import TestUtils from 'react-addons-test-utils';
+import TestUtils from 'react-dom/test-utils';
```

pavanmehta91:
is the development on the repo active? Is it usable with the new reactjs version. If no then we could fork and change it and make our own version. Otherwise contribute towards the issue by PRs.

jquense:
There is already a PR for the propTypes fix. I'll take  another one switching to the createReactClass package. These are just warnings of upcoming changes. The package is still fully compatible with react 15.5

davidgoli:
@jquense it looks like the `rm prop-types` branch got merged, but the built packages in `lib/` haven't been updated. Probably the package just needs to be built, committed, and published to npm?

bigkrp:
I also have errors in console. It seams lib folder didn't updated

akraines:
I just updated to 15.5.4 (from 15.3.0) and suddenly none of my multiselect widget's are working. 
I get the following error: 
Uncaught Error: traverseParentPath(...): Cannot traverse from and to the same ID, ``.
Is there something else that I need to change?

akraines:
NM, problem is solved. Either by adding create-react-class to the project or by removing a different offending library.

davidgoli:
@akraines ' problem may be solved, but this issue is still an issue.

davidgoli:
@jquense would you please consider rebuilding the package to `dist/` and publishing to npm? that will fix deprecation warnings for a lot of people using your package. Thanks!

ericvilleneuve:
I'm having the same issue with `dist` not having been updated.
Also, I couldn't find a PR or a branch for a fix to switch to the createReactClass package. Is there already an effort put towards this?

jquense:
no one is working on the createReactClass switch for v3 v4 is already switched over.

ericvilleneuve:
Do you think v4 is close to a release?

[...]

PR #599, opened 2017-07-03
[v4] Bump dependencies to solve React 15 deprecations

vdh:
Bumping `react-component-managers` to `^3.0.2`, and `react-overlays` to `^0.7.0`, as mentioned in #559


jquense:
thanks!

PR #619, opened 2017-08-17
Fix aria-owns referencing hidden element

Filip0:
The multiselect input field has the aria-owns attribute referencing the elements that it owns.
On initial render when the popup containing the List has yet to be rendered, so the reference to that ID is invalid.


jquense:
Thanks for the PR! would you mind making this change for the other list components (dropdown, combobox, datetimepicker)?

PR #653, opened 2017-10-03
Use transpiled index.js instead of localizer.js

balmma:
Consistency across packages and no need do special handling for react-widgets-simple-number.

### DIFF
--- a/packages/localizer-simple-number/package.json
+++ b/packages/localizer-simple-number/package.json
@@ -2,7 +2,7 @@
   "name": "react-widgets-simple-number",
   "version": "4.0.2",
   "description": "A globalize.js localizer for react-widgets",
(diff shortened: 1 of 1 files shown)


jquense:
thanks! i'll push a release a bit later

samalexander:
@jquense, when are you planning to cut a release with this fix in? It is blocking us from using the NumberPicker component (outside of switching everything to Globalize). Thanks for all your help!

balmma:
For the time being you can transpile the localizer.js file manually (even online at  https://babeljs.io/repl/) and require the transpiled file directly. After the release you just have to change it back to the module.

obrejla:
Hi, when do you think the new version will be released? Thanks.

PR #661, opened 2017-10-09
Fixed ambiguous use of mixed operators

samalexander:
Added parentheses to a function of the simple-number localizer, which uses mixed operators. This makes the developer's intentions clearer and the code more readable.


jquense:
thanks!

Issue #577, opened 2017-05-22
[4.0.0-rc.13] Multiselect vertical alignment when spanning multiple lines


huan086:
Items in Multiselect get cut off when spanning multiple lines. Default vertical align for display:table seems to be middle, causing the cutoff. Solved by adding

```css
.rw-multiselect .rw-widget-picker > * { vertical-align: top; }
```

jquense:
can you post a picture of the visual glitch you are seeing? I've tested the MS with multi line tags and they work ok.

huan086:
![multiselect](https://user-images.githubusercontent.com/1448788/27825910-bbb37692-60e3-11e7-8427-a5689e66b092.png)

jquense:
I don't think this is a problem with the default styles of the multiselect since it works in our tests

![image](https://user-images.githubusercontent.com/339286/28219415-e1add4ca-6889-11e7-9620-8e804a2e19df.png)

Perhaps you have a conflict with your styles?

huan086:
Please test in Firefox. The alignment problem doesn't occur in Chrome, and IE.

JTFlan:
Yes, this issue only exists in Firefox from what i've noticed as well and the fix @huan086 made solves it for me

dikond:
I can confirm that this issue exists in Firefox (Nightly 59.0a1).

PR #684, opened 2017-11-05
Fix Multiselect alignment in Firefox (#577).

huan086:
Add `.rw-multiselect .rw-widget-picker > div { vertical-align: top; }` style to prevent clipping of 2nd row in Firefox


dikond:
This css will screw dropdown lists. The text will be not be centered but near the top.

PR #709, opened 2017-12-17
Number Picker Conditional Rendering of Up and Down Buttons 

nklu:
Allow for the Conditional rendering of the Up and Down buttons on the NumberPicker.

"removeButtons" may not be the best of names but "disabled" seems misleading.  It should be a negative verb so it can be optionally true and backwards compatible.

Buttons could also be moved into another function if the current style is not preferred.

### DIFF
--- a/packages/react-widgets/src/NumberPicker.js
+++ b/packages/react-widgets/src/NumberPicker.js
@@ -112,6 +112,11 @@ class NumberPicker extends React.Component {
      */
     parse: PropTypes.func,
 
+    /**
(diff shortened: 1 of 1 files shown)


okb1100:
Is this going to get merged?

jquense:
One off API's like this aren't ideal in terms of consistent APIs. I am considering a "hide caret" option broadly for v5 but unsure if it will make it. For the time being hiding buttons is best done with CSS if needed

Issue #596, opened 2017-06-28
DropdownList doesn't search items above the selected value


jedwards1211:
To repro, in the demo example:

1. Click the dropdown and type in `lucile`
2. Hit enter to select `Lucile Williams`
3. Click the dropdown and type in `gilbert`
4. The dropdown should outline `Gilbert Diaz` but it doesn't
5. Type in `oscar`
6. The dropdown outlines `Oscar Diaz`

jquense:
yeah it starts searching at the focused item, it should loop back around tho

rokborf:
Any updates here?

jquense:
updates to what? the question was answered

jedwards1211:
@jquense Oh, I thought by

> it should loop back around tho

you meant this is a bug and you still need to make it loop back around.  Are you saying you believe the existing code already loops back around?  It seemed to me like it wasn't doing that.

jquense:
you're right! sorry yes that is a bug

PR #717, opened 2018-01-05
Fix DropdownList search above items issue

rokborf:
The solution for this issue:
https://github.com/jquense/react-widgets/issues/596

Because we don't need to rely on the focused or selected item to search through the whole list.

### DIFF
--- a/packages/react-widgets/src/DropdownList.js
+++ b/packages/react-widgets/src/DropdownList.js
@@ -493,8 +493,7 @@ class DropdownList extends React.Component {
       'search',
       () => {
         var list = this.list,
-          key = this.props.open ? 'focusedItem' : 'selectedItem',
-          item = list.next(this.state[key], word)
+          item = list.next(null, word)
 
         this._currentWord = ''
         if (item) cb(item)
--- a/packages/react-widgets/src/DropdownList.js
+++ b/packages/react-widgets/src/DropdownList.js
@@ -497,7 +497,7 @@ class DropdownList extends React.Component {
           item = list.next(this.state[key], word)
 
         if(item === this.state[key]) {
-          item = list.prev(this.state[key], word)
+          item = list.next(null, word)
         }
 
         this._currentWord = ''
--- a/packages/react-widgets/src/DropdownList.js
+++ b/packages/react-widgets/src/DropdownList.js
@@ -493,7 +493,12 @@ class DropdownList extends React.Component {
       'search',
       () => {
         var list = this.list,
-          item = list.next(null, word)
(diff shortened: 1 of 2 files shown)


packages/react-widgets/src/DropdownList.js
```diff
@@ -493,8 +493,7 @@ class DropdownList extends React.Component {
       'search',
       () => {
         var list = this.list,
-          key = this.props.open ? 'focusedItem' : 'selectedItem',
-          item = list.next(this.state[key], word)
+          item = list.next(null, word)
```
jquense: this isn't quite right. I think it should start from the current item and loop back around

rokborf: Do you have any concerns about starting the search from the beginning?

jquense: i think it feels less intuitive, maybe. I'm not sure, if it's reasonably easy i think i'd prefer it. At the very least it would bring back the behavior it used to have

rokborf:
hi @jquense 

do you have any concerns about my latest update?

packages/react-widgets/src/DropdownList.js
```diff
@@ -496,6 +496,10 @@ class DropdownList extends React.Component {
           key = this.props.open ? 'focusedItem' : 'selectedItem',
           item = list.next(this.state[key], word)
 
+        if(item === this.state[key]) {
+          item = list.prev(this.state[key], word)
```
jquense: So this will move the last item, then to second to last item, then to the last item, etc. etc.

what you want to do here is if the next item is the current item, start over at the beginning, e.g. `list.next(null, word)`

rokborf: No, actually.

First, it will make the search from *the current position* to *the end* of the list(as it was before).
If there is no result it will male search from *the current position* to *the beginning* of the list

It looks a little bit more efficient for me as we will not make a duplicate search on the second half of the list, but I can update PR with your suggestion if you feel it more suitable.

jquense: Did you try it? It loops on the last two matching items. I think i'd prefer that the search loops, that's the old behavior and should be what is restored, thanks!

rokborf: yeah, you're right, will update

rokborf: done, not sure about testing - as you don't nest `describe` I've added a similar test with the back direction

rokborf:
hi @jquense
can it be merged?

jquense:
thanks!

rokborf:
thank you too

PR #719, opened 2018-01-10
Disable weekends and certain dates

timneyens:
I created an update so that the weekends and certain dates can be disabled in the datepicker.

### DIFF
--- a/packages/react-widgets/src/Calendar.js
+++ b/packages/react-widgets/src/Calendar.js
@@ -210,6 +210,11 @@ const propTypes = {
 
   /** @ignore */
   tabIndex: PropTypes.any,
+
+  /**
+  * Determines if a date should be disabled
+  */
+  isDateDisabled: PropTypes.func,
 }
 
 /**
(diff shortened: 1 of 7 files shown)


jquense:
thanks for the PR. I think i'd to handle something like this more generally. Instead of an array of dates, the calendar could accept a `isDisabled` function that is called on each date as they are rendered, then the user can pick and choose any criteria they want for disabling dates

timneyens:
Thanks for the reply, I removed the previous code and added a new param 'isDateDisabled'. This is a function which checks if a date should be disabled or not.

QuantumHive:
I need this behavior, when do you plan to merge this PR?

jquense:
long time coming but can you rebase this?

timneyens:
Done ;)

Shadowman4205:
Hello, is this functionality available on main?

rishitank:
Hi, will this pull request be merged to master soon? This feature is crucial for me, thanks.

papiro:
Does anybody have a work-around to allow disabling certain dates?

papiro:
My work-around goes something like this:

```
DayComponent ({date, label}) {
  if (this.props.disableWeekends) {
    const dayOfWeek = date.getDay();
    isValidDate = dayOfWeek > 0 && dayOfWeek < 6;
  }

  return (
    <div 
      className={styles.calendarCell}
      onClick={evt => {
        !isValidDate && evt.stopPropagation();
      }}
    >
      {label}
    </div>
  )
}
```
I made the `DayComponent` a method on my DateTimePicker class so that I could reference props and state. I also added some css like so:
```
.rw-cell {
  padding: 0;
  border: none;
}

:local(.calendarCell) {
  padding: 0.25rem;
}
```
This css is so the custom DayComponent takes up the full `<td>` so no clicks escape it. Of course, you can style the disabled cells how you wish.

trevorhex:
@alejandrovillegas can this please be merged? id rather not have to rework and switch to a different component  😝

PR #721, opened 2018-01-12
Read only input

timneyens:
Added readOnlyInput prop to make input field read only.

Related: #176, #542


avinesse:
Hi, this feature would be a huge help for me, do you know if it will be merged soon?

jquense:
this was added in the most recent version via `inputProps`

PR #724, opened 2018-01-15
Add localizer-date-fns

ecancino:
Hi @jquense,

I am a little stuck in how to ensure this is correct, I can't make storybook book run the stories, seems to be loading two instances of react-dom, I will keep trying.  

If you have some time, could you review this PR and point out anything that you feel is incomplete? 

I have added the package and added the documentation to the best of my knowledge.

Thank you.

PR #736, opened 2018-01-29
Fixes storybook by using the same major version of React

marnusw:
As @jquense pointed out in #724 the stories were broken because more than one instance of React was being loaded. This brings the react major version of the `storybook` package up to date with the `react-widgets` package resolving this issue.

After running `lerna bootstrap` with this update some of the `*.js.map` files had changed. I'm not sure whether these should be included on this branch. If not the second commit can be removed.


jquense:
Thanks!

PR #743, opened 2018-02-02
Update required date formats to include weekday.

### LINKED ISSUE #734: Required Date Formats not up-to-date.
The "weekday" field is used in https://github.com/jquense/react-widgets/blob/9f156cccad056f3ebc66d4f71691cdce2e5fd1dd/packages/react-widgets/src/Month.js#L98 but is not declared as required in https://github.com/jquense/react-widgets/blob/79b3a448ebea4ac1c3d456343181e2d2eb8f4af3/packages/react-widgets/src/util/localizers.js#L12-L23 or on your site.

This breaks the DateTimePicker the minute you try to open it if you had not added a "weekday" field in your formats.

marnusw:
Closes #734

### DIFF
--- a/packages/react-widgets/src/util/localizers.js
+++ b/packages/react-widgets/src/util/localizers.js
@@ -15,6 +15,7 @@ const REQUIRED_DATE_FORMATS = [
   'time',
   'header',
   'footer',
+  'weekday',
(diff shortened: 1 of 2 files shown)


jquense:
thanks!

PR #738, opened 2018-01-29
Add localizer-date-fns with configurable locales

marnusw:
This PR supersedes #724 as it incorporates the changes merged in ecancino/react-widgets-dates#1 and will close #723. 

The implementation will work in the Storybook once #736 has been merged.

It still needs docs for the localization section before merging which I'll add soon, but in the meantime the implementation is available for comment.

PR #744, opened 2018-02-02
Fix passing the date format to localizer.parse

marnusw:
All localizers already have the signature `parse(value, format, culture)` and the documentation reflects the same which probably makes `parse.call(this, value, culture)` a bug.

This fix is required to make parsing with `localizer-date-fns` in #738 work.


marnusw:
@jquense Is anything still blocking this PR which I could help with? At the moment the `date-fns` localizer errors on parse after the text input of the DateTime picker is changed because of this issue. It would be good for us to have that resolved.

jquense:
thanks!

PR #745, opened 2018-02-02
add minSearch for MultiSelect component

itajaja:
lmk if you need extra documentation/tests 🙌

### DIFF
--- a/packages/react-widgets/src/Multiselect.js
+++ b/packages/react-widgets/src/Multiselect.js
@@ -56,7 +56,12 @@ let propTypes = {
   onChange: PropTypes.func,
 
   searchTerm: PropTypes.string,
+
+  /**
+   * The minimum number of search characters needed before the list appears.
+   */
   minSearch: PropTypes.number,
+
   /**
    * @type {function (
    *  searchTerm: ?string,
@@ -153,7 +158,7 @@ class Multiselect extends React.Component {
     filter: 'startsWith',
     value: [],
     searchTerm: '',
-    minSearch: 1,
+    minSearch: 0,
     listComponent: List,
   };
 
@@ -262,7 +267,7 @@ class Multiselect extends React.Component {
 
   handleInputChange = (e) => {
     this.search(e.target.value, e, 'input')
-    if (e.target.value.length >= this.props.minSearch) this.open()
+    this.open()
   };
 
   @widgetEditable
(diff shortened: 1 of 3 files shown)


www/language.js
```diff
@@ -144,6 +144,10 @@ module.exports = {
     The minimum number of search characters needed before filtering begins.
   `,
 
+  minSearch: d => stripIndent`
+    The minimum number of search characters needed before the list appears.
+  `,
```
jquense: I think since this is only relevant to Multiselect you can do it as a jsDoc-esque comment on the prop-type in the file

jquense:
a test would be great if you have time, otherwise it's not huge deal

packages/react-widgets/src/Multiselect.js
```diff
@@ -643,7 +648,7 @@ class Multiselect extends React.Component {
   }
 
   open() {
-    if (!this.props.open)
+    if (!this.props.open && (this.props.searchTerm.length >= this.props.minSearch))
```
jquense: im a little worried about this, we want the widget to still open if the user clicks the open button i think

itajaja: but that kind of defies the point of the min search right? how would min search work if it opens when you click, no matter how long the search term?

jquense: what if you want to just select something from the list?

jquense: the default is `1` :P

itajaja: no default is 0

jquense: sneaky

itajaja: so what do you think? :)

jquense: idk i think if you never want this to open when the search is below a x we probably don't need this, since you can control searchTerm and always not allow the input to close when its <= whatever. The case where you don't want it to open while searching i think requires the change here, since you can't easily distinguish why `onToggle` is called. But I think your usecase doesn't case why it's called

itajaja:
closing this. This is easily achievable with

```js
onToggle(open) {
  if (this.state.searchTerm < 2) return
  this.setState({ open })
}
```

Thanks 🙏

PR #754, opened 2018-02-12
Localization docs for react-widgets-date-fns.

marnusw:
This is a follow-on of #738 adding documentation to the Localization page of the documentation website.

It's not been tested since the local build of the docs gave a `WebpackOptionsValidationError` when run which I didn't investigate. The additions should be fine though.

cc @ecancino


jquense:
awesome thanks!

Issue #764, opened 2018-03-02
scss variables.scss incorrect order for @btn-bg


jafin:
Noticed in v 4.2.0 (4.2.1)

In variables.scss  
$btn-bg 

is defined after 

$state-bg-hover:            darken($btn-bg, 10%) !default;

Causing:
```
ERROR in ./~/css-loader!./~/sass-loader/lib/loader.js!./ClientApp/sass/styles.scss
Module build failed:
undefined
                                  ^
      Undefined variable: "$btn-bg".
      in c:\Data\Dev\CCSS\src\CCSS\node_modules\react-widgets\lib\scss\variables.scss (line 34, column 36)
 @ ./ClientApp/sass/styles.scss 2:14-124
```

zachGl:
Is this issue fixed because I am still experiencing it?

jquense:
i haven't put out a release in the _4 hours_ since the commit landed, no

PR #765, opened 2018-03-02
Order of btns to fix dependency for state-bg-hover  #764

jafin:
Moved order of btns before Widget States


jquense:
fixed thanks!

PR #775, opened 2018-03-26
css: correct styles to support long placeholders for Multiselect

skulden13:
Hi @jquense,

I've noticed that the Multiselect control width is too wide on mobile or e.g. when there is a lack of space and a placeholder is too long.
You could easily repeat this behavior by setting very long placeholder and resize the layout to mobile view.

That happens because of the size parameter, and the Input has property width: auto.
https://github.com/jquense/react-widgets/blob/080ed2d6130a9719459b26422c05b625f8fca3c1/packages/react-widgets/src/MultiselectInput.js#L38

As a workaround, I offer to set text-overflow for a placeholder & width: 100% for a blur state of the Multiselect input.

Thanks, Denis

### DIFF
--- a/www/src/pages/migration-v4.md
+++ b/www/src/pages/migration-v4.md
@@ -3,7 +3,7 @@
 
 In the past `react-widgets` included a set of localizers for use with popular
 I18n libraries like Globalize and moment.js. One issue with this approach is that
-we couldn't iterate on those parts of hte package without needing to bump the version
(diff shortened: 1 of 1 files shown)


GeyseR:
I've restructured ivelum/react-widgets repository, so please consider to close this PR and use https://github.com/jquense/react-widgets/pull/802 instead

skulden13:
Thank you @GeyseR!

Issue #763, opened 2018-02-20
DateTimePicker inputprops doesn't works


Gi972:
Hello,

When I put `<DateTimePicker   inputProps={{readOnly:true}} />`

It doesn't work and when I try "readonly" I have a Warning: 
"Invalid DOM property `readonly`. Did you mean `readOnly`?" and it doesn't work also.

Thanks

breadadams:
The `readOnly` prop needs to be set directly on the `DateTimePicker` @Gi972 - as specified [here](https://jquense.github.io/react-widgets/api/DateTimePicker/#readOnly). Although I had to [check the source](https://github.com/jquense/react-widgets/blob/master/packages/react-widgets/src/DateTimePicker.js#L327) to figure it out.

jquense:
I do think this should maybe work tho in cases where you want a readonly input but also have the buttons work.

breadadams:
Aha yes, I hadn't realised that the `readonly` prop handled the button too. I could address it in #785 @jquense

Gi972:
Hello guys,

Exactly readonly just for the input  and not for the both

Thanks for the fix.

PR #785, opened 2018-04-27
Update DateTimePickerInput.js

breadadams:
> Allow full control over `DateTimePickerInput`'s `<input/>` field props

This update allows the setting of `className` & `type` on the `DateTimePicker`'s input component, via the `inputProps` prop.

### DIFF
--- a/packages/react-widgets/src/DateTimePicker.js
+++ b/packages/react-widgets/src/DateTimePicker.js
@@ -363,7 +363,7 @@ class DateTimePicker extends React.Component {
         autoFocus={autoFocus}
         placeholder={placeholder}
         disabled={disabled}
-        readOnly={inputReadOnly || readOnly}
+        readOnly={inputReadOnly !== null ? inputReadOnly : readOnly}
         value={value}
         format={getFormat(this.props)}
         editFormat={editFormat}
--- a/packages/react-widgets/src/DateTimePicker.js
+++ b/packages/react-widgets/src/DateTimePicker.js
@@ -363,7 +363,7 @@ class DateTimePicker extends React.Component {
         autoFocus={autoFocus}
         placeholder={placeholder}
         disabled={disabled}
-        readOnly={readOnly || inputReadOnly}
+        readOnly={inputReadOnly || readOnly}
         value={value}
         format={getFormat(this.props)}
         editFormat={editFormat}
--- a/packages/react-widgets/src/DateTimePicker.js
+++ b/packages/react-widgets/src/DateTimePicker.js
@@ -336,7 +336,7 @@ class DateTimePicker extends React.Component {
       name,
       tabIndex,
       autoFocus,
-      inputProps,
+      inputProps = {},
       'aria-labelledby': ariaLabelledby,
       'aria-describedby': ariaDescribedby,
     } = this.props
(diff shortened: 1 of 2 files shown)


packages/react-widgets/src/DateTimePickerInput.js
```diff
@@ -62,16 +64,16 @@ class DateTimePickerInput extends React.Component {
...
     let { textValue } = this.state
 
     let props = Props.omitOwn(this)
 
     return (
       <Input
         {...props}
-        type="text"
-        className="rw-widget-input"
+        type={type}
```
jquense: i don't actually think this should be configurable, any other input type is likely to break the inputs and trigger conflicting browser behavior.

packages/react-widgets/src/DateTimePickerInput.js
```diff
@@ -62,16 +64,16 @@ class DateTimePickerInput extends React.Component {
...
+        className={`rw-widget-input ${className}`}
```
jquense: lets use `classnames` package for this like we do elsewhere to handle the null case

packages/react-widgets/src/DateTimePickerInput.js
breadadams: Good point 👍

breadadams:
I've applied your suggestions @jquense. Also included a fix for #763

packages/react-widgets/src/DateTimePicker.js
```diff
@@ -361,7 +363,7 @@ class DateTimePicker extends React.Component {
         autoFocus={autoFocus}
         placeholder={placeholder}
         disabled={disabled}
-        readOnly={readOnly}
+        readOnly={readOnly || inputReadOnly}
```
jquense: I think this should be `inputReadOnly == null ? readOnly : inputReadOnly` that way the inputReadOnly is always used if it's set to false/true

tomaszmittlener:
Any updates on this PR? :) it would be great to have this functionality.

packages/react-widgets/src/DateTimePicker.js
breadadams: Ah right @jquense, I'd done it the other way 😅. That's `inputReadOnly || readOnly` then - will adjust the PR

jquense: quick comment before you do: i think `inputReadOnly || readOnly` is not quite right, since it will fall back for `false`, the idea is you want it to always use `true` or `false` and only fallback when its `null` or `undefined` (e.g. not added)

breadadams: 🤦‍♂️ absolutely correct @jquense, my bad. Should be good to go now!

packages/react-widgets/src/DateTimePicker.js
```diff
@@ -363,7 +363,7 @@ class DateTimePicker extends React.Component {
         autoFocus={autoFocus}
         placeholder={placeholder}
         disabled={disabled}
-        readOnly={inputReadOnly || readOnly}
+        readOnly={inputReadOnly !== null ? inputReadOnly : readOnly}
```
jquense: `!=` so it gets null and `undefined`

bmight:
I saw this was closed in #806 but didn't seem to resolve the `className` issue for inputProps in `DateTimePickerInput.js`. Is there any other way to set className on the Input component. Trying to add bootstrap validation classes to the input.

PR #793, opened 2018-05-16
Fix DateTimePicker current selected time if date is not today

sators:
If a DateTimePicker has a value that is not equal to today, then the time picker does not scroll to the nearest time value.

This fix resolves that issue and a storyboard has been added to confirm the functionality.

### DIFF
--- a/packages/react-widgets/src/TimeList.js
+++ b/packages/react-widgets/src/TimeList.js
@@ -21,33 +21,13 @@ const find = (arr, fn) => {
   return null
 }
 
-function getBounds({ min, max, currentDate, value, preserveDate }) {
-  //compare just the time regradless of whether they fall on the same day
-  if (!preserveDate) {
-    value = value || currentDate || new Date()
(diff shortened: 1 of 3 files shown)


packages/react-widgets/src/TimeList.js
```diff
@@ -41,9 +42,11 @@ function getBounds({ min, max, currentDate, value, preserveDate }) {
...
     }
   }
-
-  let start = dates.today()
-  let end = dates.tomorrow()
+  let start = dates.startOf(
+      dates.merge(value, min, currentDate),
+      'minutes'
+    )
+  let end = dates.merge(dates.add(start, 1, 'day'), max, currentDate)
```
jquense: I think this breaks min/max settings, which _should_ take into account the day and compare it "now" and not allow time selection before or after the min/max datetimes

jquense:
thanks, can you please also add a unit test 👍

PR #804, opened 2018-06-02
Use simple link for remove button in MultiselectTag component

GeyseR:
Hi!
In our internal project, it's regular practice to wrap a different kind of form elements inside <label> tag to improve accessibility and improve UI. The same happens with autocomplete inputs.
For example
```
<label>
  Select what do you want
  <Multiselect items={[....]} />
</label>
```
Unfortunately, multiselect widget doesn't work well with this kind of layout, because it uses button element for remove button. After clicking on some option in the list of choices the click event is bubbling to the parent label and browser generate click event for the first button element inside label, which is remove button for the just selected option. After that selected option disappears from multiselect list of options.
Another problem - when several values already passed as props inside Multiselect component, clicking on label removes selected options one by one, as it generates clicks on removal buttons.
Here is the minimal example for reproducing browser behavior: https://jsfiddle.net/fph2hhd1/1/ (try click on the label text)
Replacing button with anchor tag is the simplest solution, but it helps.

### DIFF
--- a/packages/react-widgets/src/Button.js
+++ b/packages/react-widgets/src/Button.js
@@ -27,7 +27,7 @@ class Button extends React.Component {
       children,
       variant = 'primary',
       spinner = <Loading />,
-      component: Tag = 'button',
+      component: Tag = 'a',
       ...props
     } = this.props
 
--- a/packages/react-widgets/src/Button.js
+++ b/packages/react-widgets/src/Button.js
@@ -38,6 +38,7 @@ class Button extends React.Component {
     return (
       <Tag
         {...props}
+        role="button"
         tabIndex="-1"
         title={label}
         type={type}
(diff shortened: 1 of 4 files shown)


jquense:
thanks for the PR! I think this generally ok, but i'm a little cocerned about the semantically incorrect use of an `a` tag here. Can we also add `role="button"` as well ensuring that this continues to work when the input is disabled (since `a` can't actually be disabled)

GeyseR:
good catch, done!

jquense:
can you confirm that disabled tags still work?

GeyseR:
Hi @jquense,

Sorry for the late reply, I didn't have too much time to check my solution in detail, we just used our own published package in NPM.
You've made a good point that `a` tag might not correctly work in a disabled state, so I've fixed its behavior. Could you please take a look one more time at the PR?

GeyseR:
hi there, just an update that in our internal project we don't need this change anymore. We switched to <label for="..."> usage without wrapping control itself.
I think the issue with wrapping is still valid, but I probably couldn't provide any real use cases anymore.

PR #805, opened 2018-06-03
fix #801, update babel, move to the global babel config (required by beta.45)

### LINKED ISSUE #801: Dev environment is not configurable from scratch
Hi!
I was trying to setup dev environment for the project, and it didn't work from scratch.
My global env:
node 10.3.0
npm 6.1.0
yarn 1.7.0
I tried to run `yarn run bootstrap` command from the root dir but it failed with an error related to wrong babel setting (not supporting object spread syntax in particular)
To make it work, I had to remove yarn.lock file and add slightly modified `.babelrc.js` file to each package dir. 
Modified version of the `.babelrc.js`:
```javascript
module.exports = api => ({
  presets: [
    [
      'babel-preset-jason',
      {
        debug: false,
        runtime: false,
        modules: api.env() === 'esm' ? false : 'commonjs',
        targets: {
          browsers: ['> 1%', 'last 4 versions', 'not ie < 9'],
        },
      },
    ],
    '@babel/preset-flow',
  ],
  plugins: [
    ['@babel/plugin-proposal-decorators', { legacy: true }],
    '@babel/plugin-proposal-optional-chaining',
  ],
})
```
Did i do something wrong?

GeyseR:


packages/localizer-date-fns/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "clean": "rimraf dist/* lib/*",
     "dist": "node ../../tools/build-localizer.js",
-    "build": "babel localizer.js -o index.js",
+    "build": "babel --config-file ../../babel.config.js localizer.js -o index.js",
```
jquense: i don't think you need to explicitly provide the config if using `babel.config.js`

GeyseR: basically, you have to, because lerna runs build scripts inside an each directory, so babel doesn't know correct root dire of the package
for this particular line we can remove `--config-file` parameter, but only because in localizer.js we don't use any features which has to be transpiled by babel (you can check this by adding spread operator in localizer.js). I personally would prefer keeping this parameter to not break build after someone decided to add expiremental ECMA features in these scripts

GeyseR: btw, here is the original answer from Babel maintainer https://babeljs.slack.com/archives/C0DFJT81H/p1527961301000007

GeyseR:
Not sure why builds failed on Travis, locally everything work fine 🤔

packages/localizer-date-fns/package.json
jquense: Ah ok, If that's the case then i'd prefer to use individual `babelrc.s` that use the new `extends` keyword rather then the config option, since i sometimes run babel outside the one npm script.

GeyseR: actually, I was stuck with the second method with updating `tools/build-localizer.js` webpack config to make it work with individual babelrc files inside each package. But I can make another trym if you want

GeyseR:
fixed in #806

PR #806, opened 2018-06-05
Styling hooks

jquense:
cc @GeyseR I had to update the babel config accidentally I think this works


GeyseR:
Great! It seems working now after your updates.
What do you think about bumping babel version in package.json files?

PR #807, opened 2018-06-06
Updated string regex in lessToSass

vdh:
Replaced second quote match with "\1", so that is only matches the first quote match (so it can match a string like "foo 'bar' baz")


jquense:
thanks!

PR #810, opened 2018-06-19
Make SelectList work with falsy values

taion:
Nothing ever used the result of find except for treating it as a boolean anyway.


taion:
ping

PR #833, opened 2018-08-14
Fix selection/focus of time

msand:
Calculation of selectedItem and focusedItem was not working because compared values were for different date than currentValue


msand:
Oh, didn't notice https://github.com/jquense/react-widgets/pull/793

PR #840, opened 2018-08-31
Fix for localization link

lujeremy:
Link to localization was leading to a 404 because it was using a relative path. Changed to absolute path.

### DIFF
--- a/www/src/templates/component.js
+++ b/www/src/templates/component.js
@@ -101,7 +101,7 @@ class ComponentTemplate extends React.Component {
           {frontmatter.localized && (
             <div className="alert alert-warning">
               <i className="fa fa-exclamation-triangle" />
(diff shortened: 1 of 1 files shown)


sergii-zhuravel:
That prefix (_react-widgets_) should be added automatically during the build phase (using **-prefix-paths** is required, like - `gatsby build --prefix-paths`). With the changes proposed in the PR the documentation will not work in  local development

lujeremy:
Thanks for the response.
"You must configure a localizer to use this component!"
Not a big issue but this public link to localization is currently leading to a 404--maybe the public docs can be re-deployed with the right build parameters?

sergii-zhuravel:
@lujeremy Yes, it's possible. Do you know if this repository still has a maintainers?

PR #846, opened 2018-09-19
Upgrade localizer-date-fns defaultFormats for date-fns >=2.0.0-alpha.8

marnusw:
In [date-fns@2.0.0-alpha.8](https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#v200-alpha8) the date format patterns were changed to match the [Unicode Technical Standard](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table). This brings the `defaultFormats` of the localizer up to date.

### DIFF
--- a/packages/localizer-date-fns/localizer.js
+++ b/packages/localizer-date-fns/localizer.js
@@ -9,7 +9,7 @@ const endOfDecade = addYears(10)
 const endOfCentury = addYears(100)
 
 function getYear(date, culture, localizer) {
-  return localizer.format(date, 'YYYY', culture)
+  return localizer.format(date, 'yyyy', culture)
 }
 
 function decade(date, culture, localizer) {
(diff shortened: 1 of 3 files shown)


jquense:
hey there thanks for the PR, we already updated the formats to be inline with the format changes in alpha.8, hence the pinned version to alpha.13

marnusw:
O, I see now. They are slightly different to what I tried to copy from the Moment localizer, but they are correct according to the new patterns.

The one that threw my was `YYYY`. Capital `Y`s give you "Local week-numbering year", whereas `yyyy` will give "Calendar year". Is that also currently correct, is the intention to use "Local week-numbering year"?

jquense:
@marnusw hey sorry, I think the settings we have is what we want, but i could definitely have misunderstood the format. I hear "local" and think that's the option we want, but i am pretty unsure on the year case.

marnusw:
No worries @jquense, I just didn't want to leave this open unnecessarily, and the bulk of the PR is indeed not applicable. 

On the year topic, it breaks my brain a bit, but I think `yyyy` will always give the full year of the particular day; the `YYYY` week-numbering year might be different on the first/last days of the year if the specific day is deemed to fall in the week of the next or previous year. I'm honestly just not sure...

But this is what the [Unicode spec](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table) says: 

`yyyy`: Calendar year (numeric).
`YYYY`: Year in “Week of Year” based calendars in which the year transition occurs on a week boundary; may differ from calendar year ‘y’ near a year transition. This numeric year designation is used in conjunction with pattern character ‘w’ in the ISO year-week calendar as defined by ISO 8601, but can be used in non-Gregorian based calendar systems where week date processing is desired.

Since the calendar widget formatting doesn't use `w` anywhere I'm thinking `yyyy` might be more applicable. 

Yes! After writing all the above I tried to find some proof: If you open the calendar at the moment and zoom out to the list of years, you will see "2020" show up twice in the list. After changing the year token in the formats to `yyyy` it correctly shows 2020 only once. 

I will update this PR to make only this change.

marnusw:
PR to change date-fns year format in #849.

papiro:
Guys, not sure if this is where the issue is, but I followed the [docs](http://jquense.github.io/react-widgets/localization/), which suggest installing `2.0.0-alpha.7` and all the words in the datepicker (current year, month, etc) were all displaying the formats ("ccccc", "PPP", etc..) instead of actual words. I debugged the issue which led me into the date-fns lib and had me seeing that this locale object the react-widgets adapter was passing in lacked a `formatters` property. Needless to say, I wasn't about to try to figure out how to fix this issue, so I just thought to update date-fns to the latest alpha, and voila! Issue fixed. If I am on to something, then I think the documentation needs to be updated...

PR #864, opened 2018-10-17
Fixes #863: force reset of DatePickerInput when date is invalid

### LINKED ISSUE #863: Invalid date stays in DateTimePicker
When an invalid date is typed in the DateTimePicker, it stays displayed in the field while the parsed value is rightfully `null`
This is the opposite behaviour described in #725 

It seems to happen only if the **first** value typed is wrong. If a correct date is typed first, then changing to an invalid date will clear the field.

To reproduce, use the demo page: http://jquense.github.io/react-widgets/api/DateTimePicker/
Steps to have the field reset:
* Choose format "MMM-dd-yyyy"
* Type **Oct-55-2018** -> field resets
* Type **Oct-12-2018** and lose focus
* Type **Oct-55-2018** and lose focus -> **field resets**

Steps to have the wrong value stay:
* Choose format "MMM-dd-yyyy"
* Type **Oct-55-2018** -> field resets
* Type **Oct-55-2018** directly -> **value stays**

Nowheresly:
Fixes #863 AFAICS, this issue seems related to the DateTimePickerInput in the handleBlur method. 

The handleBlur method is in charge of propagating calls to the DateTimePicker.onChange method.
Unfortunately, the DateTimePicker.onChange will not propagate back to the DateTimePickerInput.onChange method since the DateTimePicker does not see any difference between the current value and the new value (they are both null).

This PR tries to fix this by resetting the DateTimePickerInput immediately in case of a parse failure.

### DIFF
--- a/packages/react-widgets/src/DateTimePickerInput.js
+++ b/packages/react-widgets/src/DateTimePickerInput.js
@@ -56,6 +56,10 @@ class DateTimePickerInput extends React.Component {
     if (this._needsFlush) {
       let date = parse(event.target.value)
 
+      const dateIsInvalid = event.target.value != '' && date == null
+      if (dateIsInvalid) {
+        this.setState({ textValue: '' })
+      }
       this._needsFlush = false
       onChange(date, formatDate(date, format, culture))
     }


Nowheresly:
Any chance for this PR to be reviewed by someone? Is there anything we can do to help? @jquense

racheldelman:
@jquense Is there a reason why this wasn't approved? It would be helpful if a fix was pushed...

Nowheresly:
No need to add I would really appreciate if someone could hava a look at this PR. Yet, I know, being open source maintainer is time consuming...

jquense:
thanks folks 👍

PR #891, opened 2018-12-07
Fix: Arrow keys not working on Combobox select when data is filtered

brennanpayne:
I was seeing an issue with the arrow keys not focusing on the correct option when the list was filtered.

I started to dig into this and first noticed that the tests for the arrow keys on the `ComboBox` weren't actually doing anything.

`inst.find('List li').first().is('.rw-state-focus')` simply returns a boolean and the tests were not asserting anything about the value itself.

Updating the tests to correctly assert a value:
`expect(inst.find('List li').first().is('.rw-state-focus')).to.equal(true)` shows us that the test is actually failing.

I've updated the `ComboBox` so that these tests now pass, and also pass when the list is filtered, which I don't believe it was before.

### DIFF
--- a/packages/react-widgets/src/Combobox.js
+++ b/packages/react-widgets/src/Combobox.js
@@ -149,16 +149,16 @@ class Combobox extends React.Component {
     let accessors = getAccessors(nextProps)
     const valueChanged = value !== prevState.lastValue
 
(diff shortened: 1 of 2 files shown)


jquense:
thanks!

PR #893, opened 2018-12-10
date-fns localizer ensure parse return null when invalid date

racheldelman:
I was running into an issue with the date-fns localizer when the user inserted an invalid date. This fixes the issue by checking if parseWithOptions returns 'Invalid Date' and returns null from the parse function instead.


racheldelman:
@jquense If this makes sense, can you please merge as soon as possible? Thanks!

jquense:
thanks!

PR #900, opened 2018-12-18
Fix 897: Move aria tags to correct place

nnmn:
Fixed the aria tags according to the ARIA combobox spec at https://www.w3.org/TR/wai-aria-1.1/#combobox and now the options are being read when navigating by keyboard

### DIFF
--- a/packages/react-widgets/src/DropdownList.js
+++ b/packages/react-widgets/src/DropdownList.js
@@ -399,6 +399,9 @@ class DropdownList extends React.Component {
         {filter && (
           <WidgetPicker className="rw-filter-input rw-input">
             <input
+              aria-activedescendant={open ? this.activeId : null}
+              aria-autocomplete="list"
+              aria-controls={open ? this.listId : null}
               value={searchTerm}
               className="rw-input-reset"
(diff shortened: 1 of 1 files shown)


courtsimas:
I, too, would like this merged in.

jquense:
Hi sorry for the lack of response. There is a reason why this isn't merged. The 1.1 spec isnt broadly supported and the original combox Aria tends to work on more SR's in more browsers in my experience. We've bee reevaluating this for v5 to make it better but at the moment I'm not confident this is the best change

nnmn:
1.0 specification already says that activedescendant should go to to the input text field, not listbox: https://www.w3.org/TR/wai-aria-1.0/roles#combobox Does this implementation really work on some SRs since it is not according to the actual specification?

nnmn:
@jquense would you consider merging this fix? The screen reader functionality is currently broken and I cannot see why it should not be fixed. It has been noted as broken by our testers and real users giving negative feedback

jquense:
sorry yes, Thanks for the bump

edmond-io:
@jquense would you consider to release a version with this merge commit?

PR #902, opened 2019-01-03
Fix hidden aria-owns for DateTimePicker

wmitchell-cambria:
This sets the aria-owns attribute only if the owned elements are shown on the screen. This prevents an accessibility issue (detectable by [AXE](https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/)), and it is the same issue referenced for Multiselect in #619.

This would be my first contribution to this project, so please let me know if there is anything I missed in creating this PR. We use DateTimePicker in https://github.com/ca-cwds/cans/


jquense:
thanks!

wmitchell-cambria:
Thank _you_!

Issue #875, opened 2018-11-02
Multiselect height bug in Firefox


rkashapov:
Firefox version: 63.0.1
OS: MacOS 10.14

![height bug in firefox](https://user-images.githubusercontent.com/3455996/47903694-134b5e80-de95-11e8-9b5f-ae43b86fffbb.gif)

cho-leukeleu:
I think it would be fixed if we had: 
`
.rw-multiselect-taglist {
display: inline-block;
vertical-align: 0;
}
`
I only checked it on FF.
I might look into it and make a PR for it :)

sarahsmo:
@cho-leukeleu Did you get a PR out for this?

Issue #685, opened 2017-11-05
DatePicker IE11 styles issue


vahmur:
Hello,
First of all i would like to thank you for a great library.
![rw-ie-bug](https://user-images.githubusercontent.com/15311809/32413640-87b3dd9a-c21e-11e7-8b89-3b608a890adb.jpg)
Looks like Datepicker styles are incorrect for IE11
Another widgets have issues too.
![rw-bug-2](https://user-images.githubusercontent.com/15311809/32414553-8d93d17a-c232-11e7-8795-1d15a0fd093a.jpg)

lynndylanhurley:
I was able to fix this in my project by adding the following CSS to the component container class:

```scss
.container {
  :global(.rw-select) {
    position: absolute;
    overflow: hidden;
    top: 0;
    right: 0;
    width: 60px;
  }
}
```

(I'm using component local styles)

waynebrantley:
```
/* datetime react widget on IE11 does not look right */
.rw-i.rw-i-calendar {
    padding-left: 3px;
}

.rw-widget-picker > .rw-select {
    position: absolute;
    overflow: hidden;
    top: 0;
    right: 0;
    width: 30px !important;    
}

```

jquense:
Care to PR the change for IE? We could use some `/11` hacks for IE only css?

waynebrantley:
The css I provided above works with IE and Chrome.  No hacks necessary?

jquense:
The position absolute is less flexible over all and needs to be coordinated with the padding on the outer container, i'd prefer to keep it scoped to the IE and use a more modern approach elsewhere

peterswallow:
The fix above worked fine with a Date or a Time type (top and bottom in photo), but when it was a DateTime type (middle in photo), it only displays the Date icon, rather than the Date and the Time icons.

![image](https://user-images.githubusercontent.com/1467298/47360720-15106780-d6c8-11e8-9c38-a39dc27aa746.png)

I think its related to the fixed width. If we increase the width it would look like this:
![image](https://user-images.githubusercontent.com/1467298/47360685-0033d400-d6c8-11e8-9f2b-7ce88aa3ca27.png)

Is everyone else seeing that?

If we have a fixed width, we need 30px for a single icon, but 60px for two icons.

peterswallow:
To handle the combination of one or two icons across all 3 scenarios of Date, Time **and** DateTime, I changed the above solution to have `width:auto`, which seems to work better.

[...]

PR #904, opened 2019-01-11
Fix for issue 875 Multiselect height bug in Firefox

sarahsmo:
Fix for issue #875 

Before: 
image

After:
image

### DIFF
--- a/packages/react-widgets/src/less/multiselect.less
+++ b/packages/react-widgets/src/less/multiselect.less
@@ -22,7 +22,8 @@
 
 .rw-multiselect-taglist {
   .unstyled-list();
-  display: inline;
+  display: inline-block;
+  vertical-align: 0;
   outline: none;
 }
 


cho-leukeleu:
Good that you made the PR, I forgot about it with all the 🎄 and 🎁 (happy new year by the way!).
I checked it in IE11...and could not test anything because the whole dropdown is not shown (this is unrelated to this PR).
I also checked it in Edge and found out there are some alignment problems with this solution.
screen shot 2019-01-12 at 08 52 00
It cam be kind of mitigates by using `vertical-align: middle` instead of `vertical-align: 0`
screen shot 2019-01-12 at 08 50 08

Though, as it's been suggested in another issue (https://github.com/jquense/react-widgets/issues/685#issuecomment-444747927) adding `vertical-align: middle` to the parent div might do the trick!
screen shot 2019-01-12 at 09 00 43
Actually that fix can be applied in a lot of different places in this project and it starts looking like...there could be a more general fix to this.

jquense:
thanks!

sarahsmo:
@jquense did this get added in your 4.4.9 release??

sarahsmo:
@jquense not sure if it did, don't see any changes to your demo site
image

Issue #911, opened 2019-02-11
Fix padding and height issues around dropdown options


sarahsmo:
So there was an issue before where in the Firefox browser, items were getting cut off. See issue here #875 
To fix it, I made the following changes:
![image](https://user-images.githubusercontent.com/31173144/52585186-21990700-2e02-11e9-8cec-80562fe739e3.png)

However, it seems to have caused some unwanted side effects where the items are not properly aligned in their box: 
![image](https://user-images.githubusercontent.com/31173144/52585323-6fae0a80-2e02-11e9-9ac0-3d74d4aab67a.png)

In addition, by doing the display: inline-block, the rw-input field was always at the line below the options selected. 

I played around with some styles and have the fix to resolve all of these.

PR #912, opened 2019-02-11
Fix 875 and 911

sarahsmo:
issues fixed: #911 #875 

@jquense can you please make sure that changes made in this commit are reverted: https://github.com/jquense/react-widgets/commit/d745887af4e9b11739c637e196367b062e9e8fe0

I tried to do that here but I am not sure if I was successful.

### DIFF
--- a/packages/react-widgets/src/less/multiselect.less
+++ b/packages/react-widgets/src/less/multiselect.less
@@ -22,7 +22,8 @@
 
 .rw-multiselect-taglist {
   .unstyled-list();
-  display: inline;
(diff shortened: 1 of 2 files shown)


sarahsmo:
@jquense Is this ready for merge?

jquense:
thanks!

sarahsmo:
Not sure how it works, but it looks like my revert to commit: d745887 didn't make it in? 
 
image
The vertical-align should be gone and the display should be inline.

@jquense

PR #914, opened 2019-02-18
show caret down for multiselect always

sarahsmo:
image


jquense:
Thanks for the PR, this isn't actually a bug, but the intended design.

sarahsmo:
@jquense closing this, so instead you could add a configurable option for always showing the caret.

PR #918, opened 2019-02-22
Disable autocomplete on Input

arfin:
Most modern browsers ignore "off" value for autocomplete. Instead,
the recommended solution is to put a different value for autocomplete
(see e.g. https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164).

Alternative solution could be to make this configurable in props, but seems that the intention is to always disable autocompletion

### DIFF
--- a/packages/react-widgets/src/Input.js
+++ b/packages/react-widgets/src/Input.js
@@ -29,7 +29,7 @@ function Input({
       type={type}
       ref={nodeRef}
       tabIndex={tabIndex || 0}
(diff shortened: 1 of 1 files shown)


arfin:
@jquense - any chance you could take a look at this when convenient? We'd love to be able to disable autocomplete for our date/time picking inputs, but the current implementation doesn't allow that.
Of course please let me know if I'm misunderstanding and there's another workaround we should be using to disable input.

PR #924, opened 2019-04-02
fix: Removed `isRequired` prop type validation from `renderItem` inside `VirtualList`

vdh:
Since the `VirtualList` code has logic handling a missing `renderItem` (it falls back to either `itemComponent` or `textAccessor`), the prop shouldn't be validated as `isRequired`.


jquense:
thanks!

PR #925, opened 2019-04-09
Remove Outdated Documentation

ZitRos:
Hello!

I've found these docs confusing when trying to integrate `react-widgets`. That method doesn't work anymore - I've replaced it with the valid one.


jquense:
Thanks!

PR #927, opened 2019-04-14
Fixed SASS Styling in variables.scss through LESS

laynef:


laynef:
This fixes less to sass compiles for different calculations for styles types

packages/react-widgets/src/less/variables.less
```diff
@@ -159,9 +159,9 @@
 @multiselect-tag-border:          @btn-border;
 @multiselect-tag-bg:              @gray-lighter;
 @multiselect-tag-height-pt:       1 - (@multiselect-gutter-pt * 2);
-@multiselect-tag-gutter-fallback: @input-height * @multiselect-gutter-pt;
+@multiselect-tag-gutter-fallback: ~"calc(@{input-height} * @{multiselect-gutter-pt})";
 @multiselect-tag-gutter:          ~"calc(@{multiselect-tag-gutter-fallback} - @{input-border-width})";
-@multiselect-tag-height:          @input-height * @multiselect-tag-height-pt;
+@multiselect-tag-height:          ~"calc(@{input-height} * @{multiselect-tag-height-pt})";
```
jquense: why is this needed? calc is a runtime calculation which  may not be supported by our lowest supported browsers

laynef: It doesn't compile correctly when compiled to sass, in the last PR another person experience the same problem. 

Calc is needed because the input-height and tag-height are two different types that are not compatible without calc

jquense: sorry, I realize why this solves the problem, I'm saying it's solving it in a backwards incompatible way, so we should look for a different approach

laynef: The other approach is to maintain SASS files separately. Are the LESS files breaking for you on import? Or the compile to CSS? This fixes importing the SASS files, I don't see breaks in this but I could be wrong. Could you share what errors you are seeing?

laynef: Sorry I miss understood this message.

The other approach is to maintain SASS files separately which I would prefer personally because LESS and SASS compile differently under the hood. SASS using a C/C++ compiler, that's why I prefer it over LESS. But this is a great hot fix

laynef: Also calc is already being used in that file so it doesn't change the current browser support

laynef: Sorry I miss understood your message.

Calc is currently being used in the project already which will not change browser support

laynef:
Any chance of getting this merged. It fixes SASS compatibility

packages/react-widgets/src/less/variables.less
jquense: this is still relevant, please remove the added calcs since they are not needed, thanks!

laynef: Calc is absolutely needed if you have a better solution I am open for suggestions. They are already present in the code base so browser compatibility should not be an issue. There will be two additional calc methods where they is already calcs in the codebase. It's been year that this PR is open and people are still having issues. For two simple updates, this is my personal solution. I hope others will use it in their code bases that prefer custom SASS styling.

sanjubhambhani: hi @laynef - your fix address the compile issue but I'm noticing a strange padding issue with multiselect field. attaching two screenshots below for reference, let me know if these aren't related. thanks!

Screen Shot 2020-05-08 at 5 51 12 AM
Screen Shot 2020-05-08 at 5 51 17 AM

jquense:
See my comment in the other issue, this is a bug with RW, and is easily fixable in your own sass file. Thanks for taking the time to send a PR.

packages/react-widgets/src/less/variables.less
laynef: This is a valid concern and I would love to find a solution that fixes this however I'd had too many issues with this project and the wait time for all contributions have been too long. I  personally don't use this package anymore. Closing this PR gives me solace.

jquense: guy, not to nitpick here but you also have many multi month delays in responding to my questions and feedback here.

PR #929, opened 2019-04-18
Add: render limits to List

subwai:
Use-case:
Let's say you use the Combobox. The Combobox search is fast at filtering the data even with a large array with several thousand items. However, rendering this list initially puts a lot of stress on the browser. It's not very useful to have a several thousand items long list anyway.
By adding a render limit, we can keep the search over all items while rendering only a useful amount of them.
This change is necessary to do inside `<List />` because slicing the data outside would break the search.

### DIFF
--- a/packages/react-widgets/src/List.js
+++ b/packages/react-widgets/src/List.js
@@ -34,10 +34,10 @@ const propTypes = {
   optionComponent: CustomPropTypes.elementType,
   renderItem: PropTypes.func,
   renderGroup: PropTypes.func,
+  renderLimit: PropTypes.number,
 
   focusedItem: PropTypes.any,
   selectedItem: PropTypes.any,
-  searchTerm: PropTypes.string,
 
   isDisabled: PropTypes.func.isRequired,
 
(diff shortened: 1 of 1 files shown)


subwai:
One could argue that this should be implemented right after the filter call for each component instead. Let me know what is preferable.

I guess it is possible to achieve via a filter function instead of slice, but it's way uglier:
```es6
limit = 50;
size = 0;
searchFilter = (item, search, idx) => {
    const itemLower = item.toLowerCase();
    const searchLower = search.toLowerCase();

    if (idx === 0) {
        size = 0; // reset
    }

    if (size === limit) {
        return false;
    }

    if (itemLower.indexOf(searchLower) !== -1) {
        size += 1;

        return true;
    }

    return false;
};
```

jquense:
Hey, thanks for the PR. I definitely agree with the problem your solving here, but I think the solution isn't ideal as core default in the components. In general I've stayed away from large list solutions in the base widgets for this reason, there are lots of good solutions but they are specific to ones particular app and UX choices. Instead we try and make the widgets extensible and pluggable such that you can implement this yourself if it suits your use cases. 

One example.ple of that is react-widgets-virtualized which window the lists only rendering visible items. You can make use of the same core APIs to add this approach as well in your own app, e g. Wrap the list component

subwai:
Ah I didn't realize you could overwrite the list component.
`react-widgets-virtualized` worked great & I agree it's a better solution. Thank you.

PR #932, opened 2019-05-01
fix: focusedItem logic to better handle boolean dropdownlists item values

emmilco:
Previously, DropdownLists did a straightup boolean check on `focusedItem`, causing unexpected default focus behavior in cases where an option had a falsey value and wasn't first in the list.

To recreate:
Generate a dropdown list with `false` as an option value in a position other than first. Select `false` from your list. On reload, the page will display `false` as the selected item, but default to the first option as the focused item.


jquense:
actually @emmilco can you also check that this bug isn't present in Combobox, Autocomplete, and Multiselect? It's likely also there

emmilco:
@jquense I fixed it in Multiselect and Combobox. Autocomplete doesn't have the problem.

jquense:
many thanks

PR #934, opened 2019-05-09
feat: pass data to value component

sloria:
Use case: displaying index of the currently selected value (e.g. "2 of 6") within the value component.


sloria:
I'm not passing `index` here since that can be computed by the user. But I can add that if you think it belongs here.

jquense:
up to you on the index, without it the above could be achieved with an inline function and closure like:

```jsx
data={data}
value={({ item }) => <span>{data.indexOf(item)} of {data.length}}
```

sloria:
Fair enough--I'm fine with the closure approach. I'll go with that.

PR #939, opened 2019-05-31
fix: invalid CSS selector

atomicpages:
Legacy IE 8 styles were causing CSS validation to fail on:

* Stylelint
* https://jigsaw.w3.org/css-validator/validator

Since the goal of the library is to support IE9+ I think we can safely remove thees styles (similar to my previous submit in de8bdea)

### DIFF
--- a/packages/react-widgets/src/less/mixins.less
+++ b/packages/react-widgets/src/less/mixins.less
@@ -107,11 +107,6 @@
 
 .height-calc(@total, @minus) {
   height: ~"calc(@{total} - @{minus})";
-
-  /** ie8 fallback **/
-  margin-top: ~"-@{minus}\9";
-  height: ~"@{total}\9";
-  /** --- **/
 }
 
 .make-btn-state(@hover-color: @btn-bg-hover, @active-shadow: @btn-active-box-shadow) {
--- a/packages/react-widgets/src/less/mixins.less
+++ b/packages/react-widgets/src/less/mixins.less
@@ -107,6 +107,11 @@
 
 .height-calc(@total, @minus) {
   height: ~"calc(@{total} - @{minus})";
+
+  /** ie8 fallback **/
+  margin-top: ~"-@{minus}\9";
+  height: ~"@{total}\9";
+  /** --- **/
 }
 
 .make-btn-state(@hover-color: @btn-bg-hover, @active-shadow: @btn-active-box-shadow) {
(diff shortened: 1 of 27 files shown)


packages/react-widgets/src/less/mixins.less
```diff
@@ -107,18 +107,13 @@
 
 .height-calc(@total, @minus) {
   height: ~"calc(@{total} - @{minus})";
-
-  /** ie8 fallback **/
-  margin-top: ~"-@{minus}\9";
-  height: ~"@{total}\9";
-  /** --- **/
```
jquense: why are you removing this?

packages/react-widgets/src/less/select-list.less
```diff
@@ -27,7 +27,7 @@ input.rw-select-list-input {
   left: 0;
   top: 50%;
   transform: translateY(-50%);
-  top: ~'0.1em /9';
```
jquense: this is intentional the `/9` is a hack that causes it to only be used in older IE

packages/react-widgets/src/less/mixins.less
atomicpages: Because less produces this:

```css
margin-top: -6px 9;
height: 15px 9;
```

Note: values are not specific.

packages/react-widgets/src/less/select-list.less
atomicpages: Which fails if you use stylelint. IE 8 usage is fractions of a percent. If support is really needed then maybe we can generate a separate `ie8.css`?

jquense: Honest question, why are you linting third party css? Do you also run eslint over react-widgets.js? 

Overall, I think removing existing browser coverage to appease a linter isn't a great tradeoff. Besides removing it for v4 isn't feasible as it's a breaking change. We are already bumping the min browsers for v5

jquense: which is to say, even if they are legacy, they are the current styles and we can't just remove them without marking it as breaking!

atomicpages: Sure that's fair, is there a v5 branch this can go in to?

jquense: ya it's `v5` hopefully tho they are already gone there. I may not have gotten to that tho

packages/react-widgets/src/less/mixins.less
atomicpages: Can this one go into the current branch since less is generating invalid CSS?

jquense: we should fix the hack here to `/9`

jquense:
thanks!

PR #943, opened 2019-06-16
Simple fix to #938. Opens up a question about data-flow.

### LINKED ISSUE #938: DateTimePicker - Resetting the field when an invalid date is received is not optional
To Reproduce:
1. Go to https://jquense.github.io/react-widgets/api/DateTimePicker/
2. Remove the minutes from the default date in the demo component
3. Deselect the input
4. Notice the input value is reset without an option to disable the behavior

Expected:
The option to retain invalid values will exist so we have more control over the input and validation.

I noticed this was "fixed" here: https://github.com/Nowheresly/react-widgets/commit/360b1f44d7aed875cde6d906f34dd759ff30df60.

SukantPal:
1. Fixed structure for `DateTimePicker` by moving `DateTimePicker.js` and `DateTimePickerInput.js` into a new folder `DateTimePicker` (renamed `DateTimePicker.js` to `index.js`).

2. Fixed #938 by making the `DateTimePickerInput` component revert to the previous `DateTime`, if the new `DateTime` is invalid/cannot-be-parsed; however, this questions our data-flow.

**Data-Flow problem that needs discussion:** 

* Previous data-flow: `DateTimePickerInput` would always call `onChange` whenever the user blurred the `DateTimePickerInput` component, i.e. clicked outside. However, some applications may want updates whenever the user changes the `DateTime`, without having to click outside. Due to the implementation of `DTPInput`, that use-case cannot be fulfilled.

* New data-flow: `DateTimePickerInput` still calls `onChange` whenever it is blurred; however, it won't if the `DateTime` is invalid - this allows the parent to assume that `onChange` will never be called with `null` as the `date`. However, it still poses a problem of "which" `DateTime` to revert to.

My fix makes the `DTPInput` revert to the `DateTime` as set in the props by the parent. However, the user may not want that, as I'll show with these reproduction steps:

1. Let's say that the current-date is "Jan 26, 1998 12:12:12 PM".

2. The user clicks on the `DTPInput` and edit it to be "Jan 27, ..."; but he doesn't click outside (hence, `onChange` isn't called).

3. The user now accidentally presses "s" making the text "Jan 27s, ...", which is invalid, and then clicks outside.

(NOTE: `DTPInput` is a shorthand for `DateTimePickerInput`).

**Expected:** User wants the component to revert to "Jan 27, ...".

**Actual:** Component reverts to the original "Jan 26, ...".

**Fix:** 

1. This could be fixed by calling `onChange` whenever the user edits the text (not just when loosing focus) and the new `DateTime` is valid. We could add a prop to `DateTimePicker` to prevent this behavior called `liftPartialDates` or `liftInProgressDates`.

OR

2. We could store the `previousValidDate` in the state of `DateTimePickerInput`.

### DIFF
--- a/packages/react-widgets/src/DateTimePicker/DateTimePickerInput.js
+++ b/packages/react-widgets/src/DateTimePicker/DateTimePickerInput.js
@@ -3,11 +3,15 @@ import React from 'react'
 import { polyfill as polyfillLifecycles } from 'react-lifecycles-compat'
 import { findDOMNode } from 'react-dom'
 
-import Input from './Input'
-import { date as dateLocalizer } from './util/localizers'
-import * as CustomPropTypes from './util/PropTypes'
-import * as Props from './util/Props'
+import Input from '../Input'
+import { date as dateLocalizer } from '../util/localizers'
+import * as CustomPropTypes from '../util/PropTypes'
+import * as Props from '../util/Props'
 
+/**
+ * Input that allows the date-time to be directly edited
+ * using the keyboard.
+ */
 @polyfillLifecycles
 class DateTimePickerInput extends React.Component {
   static propTypes = {
@@ -26,6 +30,13 @@ class DateTimePickerInput extends React.Component {
   }
 
   state = {}
+
+  /**
+   * To be set whenever this input's text is altered; it
+   * lifts the new DateTime to the parent by calling
+   * {@code onChange}.
+   */
+  _needFlush = false
 
   static getDerivedStateFromProps(nextProps, prevState) {
     let { value, editing, editFormat, format, culture } = nextProps
@@ -57,11 +68,20 @@ class DateTimePickerInput extends React.Component {
       let date = parse(event.target.value)
 
       const dateIsInvalid = event.target.value != '' && date == null
-      if (dateIsInvalid) {
-        this.setState({ textValue: '' })
-      }
       this._needsFlush = false
-      onChange(date, formatDate(date, format, culture))
+
+      if (dateIsInvalid) {
+        this.setState({
+          textValue: formatDate(
+            this.props.value,
+            this.props.editing && this.props.editFormat ?
+              this.props.editFormat : this.props.format,
+            this.props.culture
+          )
+        })
+      } else {
+        onChange(date, formatDate(date, format, culture))
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


jquense:
Hey there, thanks for taking the time to work on this. I think any outcome here is likely to be a breaking change so it may be best to start from the `v5` branch instead of master. 

It looks like you've identified just why I think leave-the-invalid-date behavior is hard. There isn't any good way to fire onChange when typing, without a concept of what an "incomplete" date looks like vs an "invalid" date. You need to know when it's incomplete to not clear or fire an onChange, since the user is still typing. Native datetime inputs do this, by splitting dates into `dd`:`mm`:`yy` segments, which works great but dramatically limits input to generic, fairly non local formats. On the other hand allowing any localization strategy (what RW does) means we really _can't_ know why a date is wrong b/c most i18n libs don't offer that.

So the problem is you really can't work around needing to flush the onChange at a "commit" point, e.g. when the user moves away from the input signaling they are done. That get's to the next problem tho. If you just revert to the old value the consumer has no idea that an invalid date was entered if they wanted to say offer feedback to the user that their date was bad and they need to provide a new one. Overall i don't think that's super terrible but it's not great and ideally i've love a solution that doesn't have to make that tradeoff but not sure such a space exists

SukantPal:
Hey jquense,

I'm sure you agree that making the input blank is not an acceptable solution.

We should at least show an error message.

SukantPal:
Or as #938 requests, we could retain the invalid value and display a red error message below the input.

jquense:
From my earlier comments in that issue and others you can see my general position. Showing an error is a good specific solution, but not a good one for a library to make. Input error states are extremely app specific, and really in the purview of a controlling Form library, not the input. See how differently say Material UI does this vs Bootstrap. We need to provide a good, extensible interface that allows users to implement said error if that's what we want. which is exactly my last point above.

I'm generally fine with showing the last valid value vs null, but there are equally ugly UX edge cases. Perhaps there also needs to be an `onInvalidChange`?

SukantPal:
Yeah, I could try adding the `onInvalidChange` prop, which allows the end application to specify an error message or cause a reversion to another date. It could be passed the last valid typed date too.

What do you think about that?

jquense:
sounds good to me. Please don't move the files around tho in your next iteration, i prefer the flat structure :)

SukantPal:
You sure you want the flat structure; since it is a actually confusing for new contributors. You're only exporting 5-6 components for the applications.

Well, anyways it's your final say.

Issue #970, opened 2019-10-03
React deprecation warnings


kristho91:
When using react-widgets 4.4.11 I am getting deprecation warnings in my log.

This is due to : https://github.com/jquense/uncontrollable/issues/32 .

Any way we can push this forward and fix this issue?

leebradley:
Looks like the issue was fixed and released an hour ago in the uncontrollable library 🎉

The react-widgets uncontrollable seems pretty out of date though. The package.json is still referencing uncontrollable ^5.0.0. The version of uncontrollable with the fix is 7.0.2.

PR #978, opened 2019-12-06
fix: React deprecation warnings (#970)

rypadilla:
* upgrades uncontrollable to ^7.1.1
* fixes uncontrollable imports
* fixes DateTimePicker test cases

The only potentially breaking change is that uncontrollable changes how it handles refs, which now affects all react-widget components.

### DIFF
--- a/packages/react-widgets/test/DateTimePicker-test.js
+++ b/packages/react-widgets/test/DateTimePicker-test.js
@@ -191,7 +191,8 @@ describe('DateTimePicker', () => {
   })
 
   it('should do nothing when disabled', done => {
-    let wrapper = mount(<DateTimePicker defaultValue={new Date()} disabled />)
+    let instanceRef = React.createRef()
+    let wrapper = mount(<DateTimePicker defaultValue={new Date()} disabled ref={instanceRef} />)
 
     let input = wrapper.find('.rw-input').getDOMNode()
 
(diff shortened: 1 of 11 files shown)


mikkpokk:
@jquense Will you release that fix any time soon?

rypadilla:
@jquense, will you review and merge this PR?

packages/react-widgets/test/DateTimePicker-test.js
```diff
@@ -201,13 +202,14 @@ describe('DateTimePicker', () => {
     wrapper.find('.rw-i-calendar').simulate('click')
 
     setTimeout(() => {
-      expect(wrapper.instance()._values.open).to.not.equal(true)
+      expect(instanceRef.current.props.open).to.not.equal(true)
```
jquense, line 205: I don't think these are testing the right thing anymore, `props` won't reflect the controlled values because they are stored in store in the Component. We need to test these conditions more generally since state is no longer accessible even with a ref. 

My suggestion would be to control the tested values and assert that the callbacks are not called, e.g. `onToggle` here.

jquense (commented):
Sorry @rypadilla GH is terrible for managing many repo's and once a notification falls off the notification list it's really easy to lose. I've been using the v5 beta's the last few months so haven't felt this itch

packages/react-widgets/test/DateTimePicker-test.js
rypadilla: Good call. I'll go back through the test cases and fix things up.

rypadilla: I've fixed up the test cases in DateTimePicker-test.js

They're ready for you to review. Thanks!

jquense (approved):

jquense:
thanks! will try and push out a release ASAP

rypadilla:
I'm not sure how soon you expect to be able to push out a release. If I don't see a release in a week or so, I'll ping you again. Thanks!

PR #986, opened 2020-01-16
Pass navigatePrevIcon & navigateNextIcon props to Calendar Header

### LINKED ISSUE #896: React-Widgets Calendar component: navigatePrevIcon/navigateNextIcon not functional
The `Calendar` component does not pass the `navigatePrevIcon` and `navigateNextIcon` props through to the `Header` component: https://github.com/jquense/react-widgets/blob/b468f591dbe63225ec3dfbd2576bd08d3606d6f7/packages/react-widgets/src/Calendar.js#L478-L493

This means that when the header is drawn for the calendar view, the props for `navigatePrevIcon` and `navigateNextIcon` in the Header are undefined even when the user has specified those props for the Calendar component.

dawnmist:
These props currently have no effect due to them not being passed from the Calendar down to the Header where they need to be used, resulting in the default values being used instead of the ones defined by the Calendar props.

Fixes #896.

### DIFF
--- a/packages/react-widgets/src/Calendar.js
+++ b/packages/react-widgets/src/Calendar.js
@@ -446,6 +446,8 @@ class Calendar extends React.Component {
       max,
       culture,
       tabIndex,
+      navigatePrevIcon,
+      navigateNextIcon
     } = this.props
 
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

PR #996, opened 2020-03-06
V5 with Typescript Strict

olmobrutall:
react-widgets compiles now with Typescript Strict. 💪 

I’ve written /*HACK*/ in four or five places where I could not get the types right.

Also I haven’t run any beautyfier, but there where cases where the code didn’t work without semicolons. 

I wait for your feedback!


packages/react-widgets/src/AddToListOption.js
```diff
@@ -1,33 +1,34 @@
...
-    <Listbox {...props} className="rw-list-option-create">
-      <ListOption
-        onSelect={onSelect}
-        focused={focused}
-        activeId={activeId}
-        dataItem={CREATE_OPTION}
-        selected={false}
-      >
+"use strict";
+var __rest = (this && this.__rest) || function (s, e) {
```
jquense: looks like some compiled code got in here

packages/react-widgets/src/Calendar.tsx
```diff
@@ -430,8 +427,8 @@ function Calendar({
 
   const handleKeyDown = (e: React.KeyboardEvent<HTMLDivElement>) => {
     let ctrl = e.ctrlKey || e.metaKey
-    let key = e.key
-    let direction = ARROWS_TO_DIRECTION[key]
+    let key = e.key;
+    let direction : Direction = (ARROWS_TO_DIRECTION as any)[key]
```
jquense: does `key as Direction` not work in the index signature?

packages/react-widgets/src/Calendar.tsx
```diff
@@ -444,8 +441,8 @@ function Calendar({
         e.preventDefault()
         navigate(direction)
       } else {
-        if (isRtl && OPPOSITE_DIRECTION[direction])
-          direction = OPPOSITE_DIRECTION[direction]
+        if (isRtl && (OPPOSITE_DIRECTION as any)[direction])
+          direction = (OPPOSITE_DIRECTION as any)[direction]
```
jquense: shouldn't the types just work here if its typed as a Record?

packages/react-widgets/src/DateTimePicker.tsx
```diff
@@ -173,6 +177,140 @@ const defaultProps = {
   formats: {},
 }
 
+export interface DateTimePickerProps extends Omit<WidgetProps, "min" | "max" | "value" | "onChange" | "onSelect"> {
```
jquense: this should use `DateLocalizationProps<TLocalizer>` i think, which has the correct types for `formats`, check out how Calendar has it

packages/react-widgets/src/DateTimePicker.tsx
```diff
@@ -173,6 +177,140 @@ const defaultProps = {
...
+  /**
+   * Determines how the widget parses the typed date string into a Date object. You can provide an array of formats to try,
+   * or provide a function that returns a date to handle parsing yourself. When `parse` is unspecified and
+   * the `format` prop is a `string` parse will automatically use that format as its default.
+   */
+  parse: string[] | string | ((str: string)=> Date | undefined);
+
+  tabIndex?: number;
+  'aria-labelledby'?: string;
+  'aria-describedby'?: string;
```
jquense: these shouldn't have to be defined, like data attributes i think

packages/react-widgets/src/DateTimePicker.tsx
```diff
@@ -233,10 +371,10 @@ const DateTimePicker = React.forwardRef((uncontrolledProps, outerRef) => {
     value: 'onChange',
     currentDate: 'onCurrentDateChange',
   })
-  const localizer = useLocalizer(messages, formats)
+  const localizer = useLocalizer(messages, formats as any /*HACK*/)
```
jquense: you can remove the hack with the note above about DateLocalizationProps

packages/react-widgets/src/SlideTransitionGroup.tsx
```diff
@@ -25,11 +34,17 @@ class SlideTransitionGroup extends React.Component {
...
+  isTransitioning?: boolean;
+  container: React.RefObject<HTMLDivElement>
+  current: React.ReactElement
+  flush?: boolean;
+  prev?: React.ReactElement | null;
 
-  constructor(...args) {
-    super(...args)
+  constructor(args: SlideTransitionGroupProps) {
+    super(args)
```
jquense:
```suggestion
  constructor(props: SlideTransitionGroupProps) {
    super(props)
```

packages/react-widgets/src/SlideTransitionGroup.tsx
```diff
@@ -10,13 +11,21 @@ const active = `${prefix}-active`
...
-const clone = (el, cls) =>
-  el &&
-  React.cloneElement(el, {
+const clone = (el: React.ReactElement | null | undefined, cls: string) => el &&  React.cloneElement(el, {
     className: cn(el.props.className, prefix, cls),
   })
 
-class SlideTransitionGroup extends React.Component {
+interface SlideTransitionGroupProps{
+  onTransitionEnd: (node: ChildNode, hadFocus: boolean | null)=> void;
```
jquense: where is this ChildNode  coming from?

packages/react-widgets/src/SlideDownTransition.tsx
```diff
@@ -92,18 +106,19 @@ class SlideDownTransition extends React.Component {
...
       >
-        {(status, innerProps) => (
+        {(status : TransitionStatus, innerProps: any) => (
           <div
             {...innerProps}
             className={cn(
               className,
               dropUp && 'rw-dropup',
-              transitionClasses[status],
+              (transitionClasses as any)[status],
```
jquense:
```suggestion
              transitionClasses[status as string],
```
maybe?

packages/react-widgets/src/Popup.tsx
```diff
@@ -32,29 +50,33 @@ const Popup = React.forwardRef(
...
     )
-  },
-)
+  }
+);
 
-Popup.displayName = 'Popup'
-Popup.propTypes = propTypes
-Popup.defaultProps = defaultProps
+(Popup as any).displayName = 'Popup';
```
jquense: why is this downcast needed?

packages/react-widgets/src/NumberPicker.tsx
```diff
@@ -286,7 +372,7 @@ function NumberPicker(uncontrolledProps) {
           tabIndex={tabIndex}
           value={clampedValue}
           placeholder={placeholder}
-          autoFocus={autoFocus}
+          /*autoFocus={autoFocus}*/
```
jquense: seems like a bug

packages/react-widgets/src/NumberPicker.tsx
```diff
@@ -247,26 +333,26 @@ function NumberPicker(uncontrolledProps) {
...
+    return step(-pStep!, event)
   }
 
-  function step(amount, event) {
+  function step(amount : number, event: React.KeyboardEvent<HTMLDivElement> | React.MouseEvent<HTMLButtonElement>) {
     const nextValue = (value || 0) + amount
 
     handleChange(
-      precision != null ? round(nextValue, precision) : nextValue,
+      precision != null ? round(nextValue, precision) as any as number /*HACK*/ : nextValue,
```
jquense: oooo this is just a bug

packages/react-widgets/src/NumberPicker.tsx
```diff
@@ -132,6 +132,92 @@ const defaultProps = {
   step: 1,
 }
 
+export interface NumberPicker extends Omit<React.HTMLAttributes<HTMLDivElement>, "onChange"> {
```
jquense: widget props should probably used something out of shared.ts

packages/react-widgets/src/NumberPicker.tsx
```diff
@@ -16,29 +16,29 @@ import { notify } from './util/widgetHelpers'
...
     }),
     500,
-  )
+  );
 
   return cancel
 }
 
-function clamp(value, min, max) {
+function clamp(value : number | string | null | undefined, min : number , max: number) {
```
jquense: the string shouldn't be in here, what is requiring that?

packages/react-widgets/src/NumberInput.tsx
```diff
@@ -186,9 +211,9 @@ class NumberPickerInput extends React.Component {
         className="rw-widget-input"
         onChange={this.handleChange}
         onBlur={this.handleBlur}
-        aria-valuenow={value}
-        aria-valuemin={isFinite(min) ? min : null}
-        aria-valuemax={isFinite(max) ? max : null}
+        aria-valuenow={value as any as number /*HACK*/}
```
jquense: value should aways be a number?

packages/react-widgets/src/MultiselectInput.tsx
```diff
@@ -34,9 +34,9 @@ const MultiselectInput = React.forwardRef(
       />
     )
   },
-)
+);
 
-MultiselectInput.displayName = 'MultiselectInput'
-MultiselectInput.propTypes = propTypes
+MultiselectInput.displayName = 'MultiselectInput';
+(MultiselectInput as any).propTypes = propTypes;
```
jquense: this is confusing, components should allow propTypes

packages/react-widgets/src/AddToListOption.js
olmobrutall: File removed

packages/react-widgets/src/Calendar.tsx
olmobrutall: Fixed

packages/react-widgets/src/Calendar.tsx
olmobrutall: Fixed

packages/react-widgets/src/DateTimePicker.tsx
olmobrutall: Fixed, but I had to keep the Omits for onChange and onSelect

packages/react-widgets/src/NumberPicker.tsx
olmobrutall: I added parseFloat, not sure if this is a fix

packages/react-widgets/src/NumberPicker.tsx
olmobrutall: using `export interface NumberPickerProps extends WidgetHTMLProps, Omit<WidgetProps, "onChange">` now

packages/react-widgets/src/NumberPicker.tsx
olmobrutall: the clamp function itself was testing for '', changed to `(value as any) === ''`

packages/react-widgets/src/NumberInput.tsx
olmobrutall: Yes: `'aria-valuenow'?: number;`

packages/react-widgets/src/MultiselectInput.tsx
olmobrutall: in this case is fixed.

For other cases like `NumericPicker` its more complicated, requiring changing:
`function NumberPicker(uncontrolledProps: NumberPickerProps) {`
to
`const NumberPicker : StatelessComponent<NumberPickerProps> = function NumberPicker(uncontrolledProps: NumberPickerProps) {`

then 
`const propTypes = {`
to
`const propTypes :WeakValidationMap<NumberPickerProps> = {`

I can fix `format: CustomPropTypes.numberFormat,` by using `createChainableTypeChecker<string>(`

but I don't know how to fix `localizer: PropTypes.object,` without changing it to `localizer: PropTypes.any,`, so I roll-backed.

packages/react-widgets/src/NumberInput.tsx
jquense: i mean tho when can `value` itself not be a number that you need to cast to any and back to number, is this b/c of `round`?

packages/react-widgets/src/MultiselectInput.tsx
jquense: I should also mention that in the cases of the main widgets the prop-types and defaultProps were intentionally not typed as prop-types, b/c the the inference got very complicated and ended up either breaking the generics or being wrong. I think it's fine to hack around that if that's what this is. I don't actually find the clever prop-types type matching the react types do very helpful

olmobrutall: Agree, sometimes TypeScrip definitions get crazy complicated. I prefer an any to having to get a PhD to fix a build error.

olmobrutall:
Is there anything else that needs to be done to get this merged?

PR #1002, opened 2020-03-19
className prop now forwarded to NumberInput if passed from parent

DreadfulBot:


packages/react-widgets/src/NumberInput.js
```diff
@@ -9,6 +9,7 @@ import Input from './Input'
 import * as Props from './util/Props'
 import * as CustomPropTypes from './util/PropTypes'
 import { number as numberLocalizer } from './util/localizers'
+import cs from 'classnames';
```
jquense:
```suggestion
import cn from 'classnames';
```

packages/react-widgets/src/NumberInput.js
```diff
@@ -173,15 +176,15 @@ class NumberPickerInput extends React.Component {
...
+    let { disabled, readOnly, placeholder, min, max, className } = this.props
 
     let value = this.state.stringValue
     let props = Props.omitOwn(this)
 
     return (
       <Input
         {...props}
-        className="rw-widget-input"
+        className={cs('rw-widget-input', className)}
```
jquense:
```suggestion
        className={cn(className, 'rw-widget-input')}
```

jquense:
thanks!

PR #1005, opened 2020-03-31
fix: Just use transitionEnd

taion:
@jquense i think we can just do this now? or was the 5000 padding important (v. the 5ms default in `dom-helpers`)?


taion:
prior to https://github.com/react-bootstrap/dom-helpers/pull/87, `transitionEnd` didn't actually work when transitions weren't supported. i also removed `emulateTransitionEnd` because it didn't seem like that was useful outside of just running `transitionEnd`...

seems like perhaps `transitionEnd` should take `padding` as another argument. and maybe `parseDuration` and `TRANSITION_SUPPORTED` should not be exported...

taion:
req https://github.com/react-bootstrap/dom-helpers/pull/96 if we want to keep the 5000 ms padding

jquense:
this should be fine i think

PR #1011, opened 2020-05-07
fix: undefined originalEvent on change list item

Abderrahim-WADDIZ:
input not passing the originalEvent to onChange, so it is always undefined
```
export const TestComponent = memo(() => {
  async function onComponentChange(value, { originalEvent }) {
    console.log('event :', originalEvent);
    setTouched({ [name]: true });
    await setFieldValue(name, value.id);
    onChange(value.id);
  }

  return(
      <SelectList
        data={data}
        name={name}
        textField={name}
        itemComponent={ExampleComponent}
        onChange={onComponentChange}
      />
  )
})
```


jquense:
thanks!

Issue #1023, opened 2020-06-27
DateTimePicker - use the given input id prop as is, instead of suffixing with '_input'


karthik20:
The components should be given required control expected by the rendering component. In this case, I need to associate my label to the input element of the date time but the id is not associated as given in the prop.

Snippet from the **_DateTimePicker.js_**:
`
this.inputId = instanceId(this, '_input')
`

I think the input id needs to be resolved directly from `props.id` (if provided). I can't override with 'inputProps' prop object as the this gets overriden by the `inputId`.

PR #1024, opened 2020-06-27
DateTimePicker - use the given input id prop as is, instead of suffixing with '_input

karthik20:
[DateTimePicker - use the given input id prop as is, instead of suffixing with '_input](https://github.com/jquense/react-widgets/issues/1023)

### DIFF
--- a/packages/react-widgets/src/Combobox.js
+++ b/packages/react-widgets/src/Combobox.js
@@ -310,6 +310,7 @@ class Combobox extends React.Component {
     } = this.props
     let { accessors } = this.state
     let valueItem = accessors.findOrSelf(data, value)
+    const resovledInputId = inputProps?.id || this.inputId
 
     let completeType = suggest
       ? filter
@@ -324,7 +325,7 @@ class Combobox extends React.Component {
         {...inputProps}
         role="combobox"
         name={name}
-        id={this.inputId}
+        id={resovledInputId}
         autoFocus={autoFocus}
         tabIndex={tabIndex}
         suggest={suggest}
--- a/packages/react-widgets/src/DateTimePicker.js
+++ b/packages/react-widgets/src/DateTimePicker.js
@@ -227,7 +227,7 @@ class DateTimePicker extends React.Component {
   constructor(...args) {
     super(...args)
 
-    this.inputId = instanceId(this, '_input')
+    this.inputId = this.props.id || instanceId(this, '_input')
     this.dateId = instanceId(this, '_date')
     this.listId = instanceId(this, '_listbox')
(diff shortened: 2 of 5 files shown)


packages/react-widgets/src/DateTimePicker.js
```diff
@@ -227,7 +227,7 @@ class DateTimePicker extends React.Component {
   constructor(...args) {
     super(...args)
 
-    this.inputId = instanceId(this, '_input')
+    this.inputId = this.props.id || instanceId(this, '_input')
```
jquense, line 230: this will lead to two elements having the same `id` since `props.id` is also (correctly) passed to the container element here. Instead you want to use `inputProps.id` if inputProps is provided. 

Also can you make this same change in the other picker widgets that suffer from the same issue?

karthik20: `const resovledInputId = inputProps && typeof inputProps.id === 'string' ? inputProps.id : this.inputId;`

Does this sound good? I guess the `id` prop is not exposed as a prop as per documentation. 
`
<DateTimePickerInput
        {...inputProps}
        id={resovledInputId}
        // rest of the code
        />
`

karthik20: I would probably add helper to the widgetHelpers.
`const resolvedInputId = hasValidCustomInputId(inputProps.id) ? inputProps.id : this.inputId;` and try to add across other widgets.

jquense: `inputProps?.id || createId(this)` should be fine

karthik20: Changes made so that **DateTimePicker** and **Combobox** resolves the input id in this manner. Other widgets like **Multiselect**, **NumberPicker** pass the `inputProps `directly to the input without as is.

Didn't change **Select** as this would mean, multiple ids per select option and DropdownList's input won't have the selected value and rather the search input and also its not focusable.

packages/react-widgets/src/Combobox.js
```diff
@@ -310,6 +310,7 @@ class Combobox extends React.Component {
     } = this.props
     let { accessors } = this.state
     let valueItem = accessors.findOrSelf(data, value)
+    const resovledInputId = inputProps?.id || this.inputId
```
jquense, line 313:
```suggestion
    const resolvedInputId = inputProps?.id || this.inputId
```

The logic tho still isn't quite right. the reason `inputId` is stored on the instance at all is because it's referenced in more than one place. In this case it's used in `aria-labelledby={inputId}` on the list component which would now point to a none existent id when `inputProps.id` is provided

PR #1025, opened 2020-07-08
fix parse prop

itajaja:
this matches the proptype and documentation


jquense:
looks like the docs and prop-types are actually wrong now. I think this is minor breaking change in v5, it stops trying to be clever here, since you can already just provide a function for fancy logic

itajaja:
oh. gotcha

itajaja:
what about this?

PR #1026, opened 2020-07-13
Allow users to provide a custom Popup component

clx-pwalters:
Allowing a custom Popup component provides flexibility for when you may need to Portal the List away from it's container. Overriding List in that scenario would cause the Filter and Create options to be separate from the List itself which caused display issues.


jquense:
hey thanks for PR, I think this is a good idea propbably but i need some time to think through the API and make sure it will be compatible with changes in v5, in particular the desire to use PopperJs to do positioning for menus

clx-pwalters:
Yeah, I've published these changes to a local npm repo and I'm using it to render a Portal and Popper combination as the popupComponent so the DropdownLists work properly within Modals. Let me know if there's any way I can be helpful.

PR #1030, opened 2020-08-19
fix: add guard on "key" in Widget keydown handler

c0state:
`key` can be undefined in certain circumstances. The one I saw was with Chrome autofill drop-downs. If you "arrow down" into a Chrome autofill box and then tab out of it, `key` is undefined, presumably because it was attached to the dropdown which no longer exists.


c0state:
@jquense any comment on this?

jquense (approved):

jquense:
sorry!

PR #1044, opened 2020-10-23
Fixes #1040

### LINKED ISSUE #1040: DateTimePicker throwing `Cannot read property 'handleKeyDown' of undefined`
We have recently upgraded from `4.5.0` to `4.6.0`. Once we upgraded we started getting errors:

`Uncaught TypeError: Cannot read property 'handleKeyDown' of undefined`

#### To Reproduce

1. Open a DateTimePicker by way of clicking the calendar icon
1. Try to navigate the calendar with the arrow keys
1. If you click the time icon and use the arrow keys it behaves fine. The problem seems to be with accessing `calRef.inner`

---

The offending line from React Widgets:

```
else if (open) {
    if (open === 'date') _this5.calRef.inner.handleKeyDown(e); <-----------------------
    if (open === 'time') _this5.timeRef.handleKeyDown(e);
}
```

The props we are using:

```
<DateTimePicker
        name={name}
        onChange={onChange}
        disabled={disabled}
        min={min}
        max={max}
        value={value}
        currentDate={effectiveCurrentDate}
        onCurrentDateChange={setCurrentDate}
        autoFocus={autoFocus}
        defaultOpen={defaultOpen ? 'date' : false}
/>
```

khanusin:
Fixes issue: Opening a date picker, then using arrow keys to navigate, produces error.

### DIFF
--- a/packages/react-widgets/src/DateTimePicker.js
+++ b/packages/react-widgets/src/DateTimePicker.js
@@ -286,7 +286,7 @@ class DateTimePicker extends React.Component {
         this.close()
       }
     } else if (open) {
-      if (open === 'date') this.calRef.inner.handleKeyDown(e)
+      if (open === 'date') this.calRef.handleKeyDown(e)
       if (open === 'time') this.timeRef.handleKeyDown(e)
     }
   }


jquense (approved):

aedryan:
Can we expect a 4.6.1 release with this fix?

PR #1059, opened 2021-01-19
Fix inputProps not being applied on DropdownList

### LINKED ISSUE #963: inputProps doesn't work on DropdownList component
When passing `inputProps` to `DropdownList` props do not appear on input element.
As I see `inputProps` value-pairs are passed then to `DropdownListItem`, but then it's considered by `DropdownListItem`

dmason30:
Fixes #963

The `inputProps` were not being applied to the input in the `DropdownListInput` component.


DaneSirois:
Bumping this as I also need this fix

PR #1075, opened 2021-04-27
fix: user deletes when there are no tags in Multiselect

AaronBuxbaum:
If a user presses delete when there are no tags in the list in a Multiselect component, it will trigger a change, which may trigger unexpected handlers. Typically, changes to a component where no state changes (ie. deleting nothing) results in no changes, therefore I propose adding a check when a user tries to delete in order to guard from this case.

I also provide a simple utility function for tag list size -- the main purpose of this is to make it more clear to code readers as to why that check exists without having to understand the internals of `FocusList` -- as opposed to a direct usage like `tagList.get()[0].length`


jquense (approved):
Nice!

PR #1079, opened 2021-06-03
Fixed Combobox's inputProps' readOnly being overwritten by Combobox's readOnly

50jeffng:
### WHAT:
Following `DatePicker` component way of setting `input` `readOnly`, it fixes `Combobox` component's `input` `readOnly` being replaced by the component's `readOnly`

### REMARKS:
I'm not really the best in Typescript so I need some help of solving the `inputProp`'s type, in `DatePicker` component it has a separate file for `input` but in this I think it's unnecessary so I temporarily bypass it by ignoring it (by `ts-ignore`)


jquense (approved):

jquense:
thanks!

Issue #1089, opened 2021-07-14
DatePicker: when min is specified, and hour and/or minute of specified min is > 10, users cannot choose a different time on the same day


jyaoma:
#### Steps to reproduce
1. Add the following story to packages/react-widgets/stories/DatePicker.tsx
```js
export const min_withTimePicker = () => {
  const todayAt10AmUTC = new Date()
    .toISOString()
    .substring(0, 11)
    .concat('10:10:00.000Z')

  const todayAt10AmLocal = new Date(
    new Date(todayAt10AmUTC).getTime() +
      new Date().getTimezoneOffset() * 60 * 1000,
  )

  return <DatePicker includeTime min={todayAt10AmLocal} />
}
```
2. In the story, try to select the same date (today) at any time after 10:00, say 23:59.

#### What should happen
I should be able to type in 23:59 into the time picker field

#### What happens
I am not able to type in 23:59 into the time picker field

I believe that this is because when typing in the time, if you type in the first digit (2), it'll think that 2 is less than 10, and it'll be an invalid time, which prevents the change in the field from happening.

PR #1091, opened 2021-07-19
Fix time picker when min is specified in date picker

jyaoma:
This attempts to fix the issue that I've recently reported as an issue: https://github.com/jquense/react-widgets/issues/1089

I've implemented a new concept to `TimeInput` called `RawTimeParts` which is just the string version of `TimeParts` and made use of `RawTimeParts` when working directly with the input fields and handling change events.

Feel free to make changes to this or you can let me know if there are any issues!

### DIFF
--- a/packages/react-widgets/src/DateTimePartInput.tsx
+++ b/packages/react-widgets/src/DateTimePartInput.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 
 const padStart = (
   value: string | number | null | undefined,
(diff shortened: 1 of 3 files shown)


jquense:
hey thanks for the PR. I'm short on time at the moment so haven't been able to look over this!

ryanore:
@jquense, I appreciate the the library and respect your time, of course.
Do you have an ETA of when we could expect you to get around to this?

jquense:
ok sorry folks i keep trying to find time to look at this. I've just had a chance to do a once over this. I think this may be too complicated for a fix, it adds a lot of code and is hard to follow. I don't think that's a reflection on quality, the problem is inherently complex. The underlying issue here is that `TimeInput` isn't actually supposed to accept a min or max...and given how complex a solution to that is i'm inclined to leave it that way, which is actually how native inputs work as well.

PR #1094, opened 2021-08-09
Clear on backspace

AaronBuxbaum:


packages/react-widgets/src/scss/widget.scss
```diff
@@ -58,6 +58,10 @@
   justify-content: $picker-btn-justify-content;
   padding-left: $picker-btn-padding-left;
   padding-right: $picker-btn-padding-right;
+
+  &:not(:last-child) {
```
AaronBuxbaum, line 62: not loving the way this looks... figured it was better than creating different CSS classes for all different possible combinations of buttons, but would be up for a better idea

packages/react-widgets/src/DropdownList.tsx
```diff
@@ -391,9 +401,17 @@ const DropdownListImpl: DropdownList = React.forwardRef(function DropdownList<
...
 
       list.focus(list.prev())
+    } else if (key === 'Backspace' && allowClear && currentValue && !currentSearch) {
+      handleClear(e);
     }
   }
 
+  const handleClear = (e: React.MouseEvent<HTMLSpanElement> | React.KeyboardEvent<HTMLDivElement>) => {
+    e.preventDefault();
+    e.stopPropagation();
```
jquense, line 411: is this to prevent toggling the dropdown open as well?

AaronBuxbaum: exactly, yes

packages/react-widgets/src/scss/widget.scss
jquense: I would add a dropdown specific class for the wrapping div (to remove the inline flex style), you could handle this there.

PR #1098, opened 2021-08-18
feat: new addon Tailwind styles

jquense:


colinramsay:
Is there any documentation on how to use Tailwind support? Thanks!

jquense:
It's a tailwind plugin, include it in the plugins of your theme and don't include the original css/sass file

PR #1102, opened 2021-09-20
feat: adds a busy spinner to combobox even when caret is hidden

brian-bfly:
image

### DIFF
--- a/packages/react-widgets/src/Combobox.tsx
+++ b/packages/react-widgets/src/Combobox.tsx
@@ -407,7 +407,7 @@ const ComboboxImpl: Combobox = React.forwardRef(function Combobox<TDataItem>(
   let popupOpen = currentOpen && (!hideEmptyPopup || !!data.length)
   let inputReadOnly =
     // @ts-ignore
-    inputProps?.readOnly != null ? inputProps?.readOnly : readOnly;
+    inputProps?.readOnly != null ? inputProps?.readOnly : readOnly
 
   const Addon = () => {
     if (!hideCaret) {
@@ -421,15 +421,15 @@ const ComboboxImpl: Combobox = React.forwardRef(function Combobox<TDataItem>(
           // FIXME
           label={messages.openCombobox()}
         />
-      );
+      )
     } else if (busy) {
       return (
         <span aria-hidden="true" className="rw-btn rw-picker-caret">
           {busySpinner || Spinner}
         </span>
-      );
+      )
     }
-    return <></>;
+    return <></>
   }
 
   return (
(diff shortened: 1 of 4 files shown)


packages/react-widgets/src/Combobox.tsx
```diff
@@ -405,9 +406,32 @@ const ComboboxImpl: Combobox = React.forwardRef(function Combobox<TDataItem>(
 
   let popupOpen = currentOpen && (!hideEmptyPopup || !!data.length)
   let inputReadOnly =
-  // @ts-ignore
-      inputProps?.readOnly != null ? inputProps?.readOnly : readOnly;
-      
+    // @ts-ignore
+    inputProps?.readOnly != null ? inputProps?.readOnly : readOnly
+
+  const Addon = () => {
```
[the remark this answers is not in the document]
brian-bfly: Done.

www/plugins/examples/busy.js
```diff
@@ -14,6 +14,8 @@ module.exports = function (widgetName) {
     <>
       <${widgetName} busy />
       <br/>
+      <${widgetName} busy hideCaret />
```
jquense, line 17: I believe this example is used by multiple components, not all of which have "hideCaret" as a prop. We'd want to make a combobox specific one, or probably better make this part of a `hideCaret` example instead.

jquense:
image

Notice that the input box shadow isn't extending underneath the caret. This is because the input has the shadow and is sharing space with the caret box. There are probably a few ways to address this, but probably the easiest is to add the `rw-widget-input` to the container div when `hideCaret` is true so the box shadow is around everything

packages/react-widgets/src/Combobox.tsx
```diff
@@ -430,7 +453,7 @@ const ComboboxImpl: Combobox = React.forwardRef(function Combobox<TDataItem>(
           className={cn(
             // @ts-ignore
             inputProps && inputProps.className,
-            'rw-widget-input rw-combobox-input',
+            'rw-combobox-input',
```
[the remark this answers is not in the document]
brian-bfly: Done.

packages/react-widgets/src/WidgetPicker.tsx
```diff
@@ -4,16 +4,17 @@ import React from 'react'
...
+}: React.HTMLProps<HTMLDivElement> & { hideCaret?: boolean, busy?: boolean }) {
   return (
     <div
       {...props}
       className={cn(
         className,
         'rw-widget-picker',
         'rw-widget-container',
-        hideCaret && 'rw-hide-caret',
+        (hideCaret && !busy) && 'rw-hide-caret',
```
jquense, line 17: last thing!, i wouldn't conflate the meaning here as this isn't what you'd expect in other components. We can do this logic in Combobox where we are synthesizing the hide meaning more specifically

www/plugins/examples/busy.js
jquense: revert this?

brian-bfly: Done.

jquense (approved):

PR #1114, opened 2021-12-13
Add defaultValue type to Timeinput compoenent

danielmackenzie8:
Typescript was throwing a type error when trying to provide a defaultValue to the TimeInput component. Have defined the type mirroring the approach taken for the DatePicker component.


leafyshark:
Screenshot 2023-01-03 at 17 50 29

PR #1120, opened 2022-01-31
RW input icon button hidden fix

### LINKED ISSUE #1119: Some input contol/button icons not visible on Firefox
The Datepicker, NumberPicker and Combobox component icon buttons seem to overflow out of visibility on Firefox Dev Edition (works fine on Chrome). Can someone confirm? At first glance it seems that the overflow happens when the component width <= 260px.

(Screenshot from [homepage](https://jquense.github.io/react-widgets/) where container width is 240px, which causes overflow within component). 
image

govindsartaj:
Fixes #1119 

Adds ```width: inherit;``` for ```.rw-widget-input``` class which stops the input icon button on Datepicker, NumberPicker and Combobox components from overflowing/getting hidden when width of component is below than some threshold (approx. 250px). Thanks!

### DIFF
--- a/packages/react-widgets/src/scss/widget.scss
+++ b/packages/react-widgets/src/scss/widget.scss
@@ -184,6 +184,7 @@ fieldset[disabled] .rw-widget {
   background-color: $input-bg;
   box-shadow: $input-box-shadow;
   background-clip: $widget-background-clip;
+  width: inherit;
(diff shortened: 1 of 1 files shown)


jquense (approved):

jquense:
thanks!

PR #1140, opened 2023-01-04
try adding nodeRef

### LINKED ISSUE #1137: findDOMNode is deprecated in StrictMode - error
Im getting this error with the combobox. 

react_devtools_backend.js:4026 Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition2 which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here:

image

quinn:
attempt to fix #1137

### DIFF
--- a/packages/react-widgets/src/SlideDownTransition.tsx
+++ b/packages/react-widgets/src/SlideDownTransition.tsx
@@ -39,6 +39,13 @@ class SlideDownTransition extends React.Component<SlideDownTransitionProps> {
     onEntered: PropTypes.func,
   }
 
+  private readonly nodeRef: React.RefObject<any>
+
+  constructor(props: SlideDownTransitionProps) {
+    super(props);
+    this.nodeRef = React.createRef();
+  }
+
   getHeight(container: HTMLElement) {
     let content = container.firstChild as HTMLElement
     let margin =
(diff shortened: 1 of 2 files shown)


quinn:
@jquense looks like I need to refactor because its a class-based component.

bt:
Is `useRef` the same as `React.createRef`?

You can use `createRef` in class-based components.

quinn:
@bt yes, it is the same. I updated to use createRef. @jquense what is a good way to test this out?

jquense:
Try running the docs but in strict mode?

bt:
@quinn Did you have any luck?

astralmedia:
Any movement on this? Would love to clean that last error up on my logs :)

quinn:
@astralmedia PR should work, I haven't had a chance to run the docs in strict mode as advised

quinn:
also, I can't see why the build failed? says, expired

quinn:
getting this error when I try to run docs:

```
node:internal/crypto/hash:71
  this[kHandle] = new _Hash(algorithm, xofLen);
                  ^

Error: error:0308010C:digital envelope routines::unsupported
```

grobux:
Hi, can this PR be merged or does it needs more work?
I face the issue it should be fixing

jquense:
I've incorporated these changes in v5.8.5

PR #1159, opened 2025-01-07
Blur Input Field Automatically After Date Selection in DatePicker

asinghka:
This change should ensure that the input field in the `DatePicker` loses focus automatically after a date is selected. On devices with on-screen keyboards, it can be annoying when the input field remains focused, causing the keyboard to open unnecessarily after the user has already selected a date.

Note: **This change has not been tested at all and should be thoroughly verified before merging.**

### DIFF
--- a/packages/react-widgets/src/DatePicker.tsx
+++ b/packages/react-widgets/src/DatePicker.tsx
@@ -468,7 +468,7 @@ const DatePicker = React.forwardRef(
 
       notify(onSelect, [dateTime, dateStr])
       handleChange(dateTime, dateStr, true)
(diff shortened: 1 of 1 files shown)


jquense:
Hi there, thanks for PR but this isn't an acceptable solution to your problem. focusably dropping focus from inputs is hostile to keyboard and screen reader users. A better option for you would be to set `readOnly` on the input for mobile screens so that the keyboard doesn't open at all and users can just use the calendar picker

asinghka:
I understand the issue with my solution.

Setting the input to readOnly did the trick. Thank you so much!

